PR #37, opened 2016-12-05
Updated installation_windows_genric

KukkadapuSushma:


daminisatya:
@KukkadapuSushma Is this tested thoroughly? since it is on windows

KukkadapuSushma:
Yes I have and @samsruthi also had a look at that.

docs/installation_windows_generic.md
```diff
@@ -25,10 +25,15 @@ Install in terminal:
...
-    > git checkout master
-    > ant jar
-    > java -jar dist/susiserver.jar
+    
+    	> ant jar
+    	> java -jar dist/susiserver.jar 
+	
+    (if on running above command gives error, try to change the branch either to "master" or "development" by following commands) :
     > git checkout development
+    > git checkout master
```
KukkadapuSushma: yeah that also works the same way! And yes, let's stick to one then! @samsruti

daminisatya:
Thank you for the patch!

Issue #25, opened 2016-11-19
Change Susi apps to account and adopt new logic


mariobehling:
* [ ] Change api.asksusi.com to "Home" and use slug "/" instead of web URL
* [ ] Only show "Account Apps"
* [ ] Change top right menu "Apps" to "Account"
* [ ] Take out slug "/apps/". Instead of http://api.asksusi.com/apps/signup/ change to http://api.asksusi.com/signup/
* [ ] On apps page change background from black to white and take out effects (e.g. moving in of apps)
* [ ] Make accounting options (accounts apps) into a simple list of items (instead of app boxes)
* [x] Delete apps: 1) Boilerplate, 2) One-Click Deployment, 3) Web Scraper
* [ ] Take out left sidebar "Categories"

KukkadapuSushma:
Hello, I would like to work on this issue!

daminisatya:
@KukkadapuSushma you work now, any help you can follow up over here or on gitter

KukkadapuSushma:
Sure, thanks @daminisatya

samsruti:
@daminisatya I would like to work on this task.

jig08:
@samsruti Sure, go ahead.

@KukkadapuSushma Any updates on this ?

samsruti:
@mariobehling What are the 'Account Apps'? Do you mean to say that the Apps that user uploaded by himself/herself, i.e The personal apps?

KukkadapuSushma:
@jig08 yes I have started working on it.

samsruti:
@KukkadapuSushma what are the implementations you have done yet. So that I can start my work excluding those?

daminisatya:
@samsruti May be let's divide the work! :) 
You take up the first four and @KukkadapuSushma will take up the last four. I guess she started with deleting the apps. 

Good to see the collaborative work :)

samsruti:
@daminisatya I am stuck in here: Take out slug "/apps/". Instead of http://api.asksusi.com/apps/signup/ change to http://api.asksusi.com/signup/

This could be by shifting the apps into the /html/ directory. So I would like if you can assign this to someone better. Since I got the top 4 i could fixed only 2 :)
Here is the pull request:  https://github.com/fossasia/susi_server/pull/36

KukkadapuSushma:
I am done with partially half of my tasks, would deploy to development once after i finish completely @daminisatya

KukkadapuSushma:
[...]

PR #38, opened 2016-12-06
[Partial Fix] Change Susi apps to account and adopt new logic

KukkadapuSushma:
Under issue #25 -- removed "Categories" and "animated effect" and "deleted apps"


daminisatya:
@mariobehling Need your help here, do we need to remove the left side bar completely? Or just the category name?

mariobehling:
@KukkadapuSushma Thanks a lot. An issue, that prevents a number of people from continuously testing is that the automatic deployment still does not work for the susi server. Could you help with making it work for Heroku? https://susi-server.herokuapp.com/

KukkadapuSushma:
@mariobehling Sure, i will go through it.

PR #40, opened 2016-12-09
Fixes #39 and Lot of cleanups and refactoring

### LINKED ISSUE #39: Universal Build System, Platform Independent Setup Instructions and Repository Cleanup
@mariobehling @daminisatya @Orbiter 
I am working on to migrate the whole project build from ant to gradle based solution completely which would simplify lot of things for environment setup. Instead of having separate setup instructions for Windows, Mac and Linux, we can have a universal system, which works on any environment with least setup, all the user needs is gradle and jdk, thats it...
then the user would need to run only two commands on any environment..
gradle build && gradle run
 and all the dependencies would also be resolved through maven which can be configured through gradle... which would further reduce the repository size.., where as the current solution has many jar files which are being referenced through gradle and ant build file...
and the jar files alone occupies at least 30mb.... and the setup instructions are also too fragmented and not very clear, after implementing this it would reduce the repository size, and also the builds will be platform independent and easy to set up.

I would be able to complete this task in a week or two if everything goes well...

vivekkiran:
Fixes #39


vivekkiran:
The build is successful travis, showing it failed because of hard coded value in https://github.com/fossasia/susi_server/blob/development/test/org/loklak/tools/storage/JsonRandomAccessFileTest.java testRead() method  " File f = new File("data/accounts/own/users_201508_74114447.txt");
"

vivekkiran:
@mariobehling @Orbiter @daminisatya Please test and verify this, i have tested it throughly on my machine and its working.

mariobehling:
@vivekkiran Thank you. How could we get the Travis build work nevertheless? If the build does not pass we will not have the automatic deployment working.

vivekkiran:
@mariobehling for now I have excluded test task, so now it will build the code but wont run the test cases directly, only on running 'gradle test' it runs, This is a temporary solution, until we have test cases in a good shape.

vivekkiran:
@Orbiter @mariobehling @daminisatya  It would be nice if someone better can take up the task of writing unit/integration tests after its merged, so I would be able to fix other issues. :)

PR #50, opened 2016-12-21
remove theme 'susi' parameter

championswimmer:
if we're using own theme, we should leave theme blank.
the theme parameter should have a valid Github jekyll theme, or be blank

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,1 +1,1 @@
-theme: susi
+theme: 


Abhi2424shek:
You are correct

BrainBuzzer:
Is this page still working? I kept it as it is because it was not working without that parameter.

mariobehling:
The links on that page are not working. Please check.

PR #86, opened 2017-02-13
Adding support for untracked files in gh-pagesUpdater.sh

Abhi2424shek:
Untracked files where ommitted and this caused the bot not to push back, added support for untracked files.
In this PR: I,
- Squashed all my commits
- Included support for the untracked files

I didn't reference an issue number because there is still no issue resent for this. And the tweak is simple.
@mariobehling pls review. Thankyou.


Abhi2424shek:
Thanks @daminisatya

PR #114, opened 2017-04-19
Transition Loklak to Susi

tabesin:
Eliminate hard coded Loklak reference. Change notification mail reference to SUSI.


mariobehling:
Thanks! Please open an issue in future and match it with a pull request.

PR #121, opened 2017-04-24
Fix links in README.rst

### LINKED ISSUE #120: Links not correct on gh-pages live page
Some links are giving 404s and the logo is not showing up.

Please fix the links by updating the way the gh-pages branch is build. Please ensure the links are set in a generic way.

pri22296:
Fixes #120

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -104,26 +104,26 @@ How do I install Susi with Docker?
 ----------------------------------
 
 To install Susi with Docker please refer to the `Susi Docker
-installation readme </docs/installation_docker.md>`__.
+installation readme <installation_docker.html>`__.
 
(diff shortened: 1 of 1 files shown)


pri22296:
Those links were giving a 404 becaause they referred to markdown files which only exists in the development branch. Their corresponding html files exists in github-pages branch. So the links need to be fixed to refer the correct html. I don't think it warrants a change in the way how docs are built. It was just a case of incorrect links. Also on further testing, Making the links absolute were not necessary. So I fixed that.

mariobehling:
> I don't think it warrants a change in the way how docs are built.

Yes, it does. We need a versatile solution that works whereever we want to deploy the documentation.

PR #128, opened 2017-05-02
Include documentation files from root while fixing relative links for sphinx documentation

### LINKED ISSUE #120: Links not correct on gh-pages live page

pri22296:
Fixes #120

### DIFF
--- a/.utility/push-docs-to-gh-pages.sh
+++ b/.utility/push-docs-to-gh-pages.sh
@@ -17,9 +17,15 @@ echo -e "Installing requirements...\n"
 cd docs
 pip3 install -r requirements.txt
 
+echo -e "Including README.rst from root to docs directory"
+
+python3 include.py README.rst
+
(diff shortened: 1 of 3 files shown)


yukiisbored:
@pri22296 Can you give us a comparison of the file before and after?

pri22296:
@yukiisbored  you can check out the comparison report in the following link

https://www.diffnow.com/?report=wn3mr

yukiisbored:
@pri22296 Since Sphinx doesn't require `.html` for linking other articles. Just remove the `.html` suffix. Also as a suggestion, please make your code more clear of what it does. It should be readable without digging in.

pri22296:
@yukiisbored I have removed the .html suffix from the links. I have also added some docstrings/comments in my code wherever applicable. Thanks for your suggestions.

Issue #113, opened 2017-04-18
Add public transit to Susi AI


Fluepke:
Many public transit providers have open data portals offering realtime data. (E.g. for Berlin see [this](http://www.vbb.de/de/article/fahrplan/webservices/schnittstellen-fuer-webentwickler/5070.html)).
We should access this data directly or via a wrapper service to provide Susi users information about different means of transportation
Maybe the sourcecode from [transportr](https://github.com/grote/Transportr) could be used to achieve this.

Fluepke:
@mariobehling suggested to implement this skill via [dream.asksusi.com](https://github.com/fossasia/susi_server/blob/development/docs/susi_skill_development_tutorial.md)
This would require to have an external API, that can provide the necessary data.
Another idea would be to implement this via a ConsoleService, but it might be problematic to include and distribute the necessary API keys.

mariobehling:
@Fluepke Let's do the lightweight approach first. "dream.asksusi.com" is not intended to be a way for implementation. The idea is to provide a setting to write skills and test them right away. Currently skills need to be added to the repository. We are planning to provide an easier way, but this is out of scope of this issue.

We really need developers to write skills and understand the flow. So, for the moment using external APIs is the best approach and it moves the project further a lot. Please proceed.

sudheesh001:
@Fluepke This definitely sounds like an interesting thing to do. The intention with susi is that others who might want to use this data can also integrate it. We can always use a whitelisted IP Address of the server to ensure that the API can be used between susi and the required service by distributing and installing a key on the servers but the intention is to work with open data.

Providing API ingestion interfaces is definitely something that needs to be implemented into susi server but for starters we can always implement this as a susi dream or a consoleService and use the `conf` to store the API credentials or we could as well build an API manager.

Orbiter:
This issue shows that one of the core requirements for a public storage place must be a concept to hide application keys.  I added that idea to https://github.com/fossasia/susi_server/issues/106#issuecomment-294828643

Fluepke:
Ok, so I've tinkered around with enabler. It's working quite well and for Berlin / Brandenburg it does not require any API keys as it just parses a public available interface (is this legal)?
Whats the way to go for implementing a ConsoleService for this? Just copy the necessary code from enabler (while contributing the authors) and writing a ConsoleService to interface with enabler?

[...]

PR #176, opened 2017-05-13
Added first support for public transit API #113

Fluepke:
Pretty fast hack to get public transit up and running with susi.ai (currently at a hackathon)
Fixes: #113


dynamitechetan:
This looks good but I think this can be done using some API like wolfram aplha in a SUSI dream. Also if we do it in a SUSI dream then future updates will be easier. 
I tried something similar just now http://dream.susi.ai/p/durationbetweenplaces

saurabhjn76:
looks good, but as @dynamitechetan suggested a lightweight approach using SUSI dream would be easier to handle.

Fluepke:
This is about supporting public transit (not some Maps API for driving with car) to susi, which is not possible as a susi dream. German railroad services do not offer a nice API for this. You need to parse ugly XML.
Actually, I was asked to push this, even though it's a hackathon project and not really ready so far (no testing, etc.)
I'd be fine if this gets rejected, so it can be pushed, when it's ready :)

DravitLochan:
Nice Idea. I too was thinking to add a skill something similar. Please look for the merge conflicts

Issue #206, opened 2017-05-30
Servlet for Changing User Roles


dynamitechetan:
If the current user logged in is the admin, then he should be able to change the user roles of other users. 
For example, if admin wants to make someone a Privileged user or Change someone's role to Admin.

PR #207, opened 2017-05-30
Adding Servlet to Change User Roles

dynamitechetan:
This is a servlet that is used to change the role of any user
This requires a to be logged in.
can be tested locally on http://127.0.0.1:4000/aaa/changeRoles

Fixes issue #206

### DIFF
--- a/src/ai/susi/server/api/aaa/ChangeUserRoles.java
+++ b/src/ai/susi/server/api/aaa/ChangeUserRoles.java
@@ -1,0 +1,112 @@
+package ai.susi.server.api.aaa;
+
+import ai.susi.DAO;
+import ai.susi.SusiServer;
+import ai.susi.json.JsonObjectWithDefault;
+import ai.susi.server.*;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import javax.servlet.Servlet;
+import javax.servlet.http.HttpServletResponse;
+import java.lang.reflect.Constructor;
+
+/**
+ * Created by chetankaushik on 30/05/17.
+ * This is a servlet that is used to change the role of any user
+ * This requires Admin login
+ * http://127.0.0.1:4000/aaa/changeRoles this accepts two post parameters-
+ *  1. user    --> The username of the user whose role is to be modified
+ *  2. role    --> the new role of the user
+ */
+
+public class ChangeUserRoles extends AbstractAPIHandler implements APIHandler {
+    private static final long serialVersionUID = -1432553481906185711L;
+
+    @Override
+    public BaseUserRole getMinimalBaseUserRole() { return BaseUserRole.ADMIN; }
+
+    @Override
+    public JSONObject getDefaultPermissions(BaseUserRole baseUserRole) {
+        return null;
+    }
+
+    @Override
+    public String getAPIPath() {
+        return "/aaa/changeRoles";
+    }
+
+
+    @Override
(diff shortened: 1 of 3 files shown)


src/ai/susi/server/api/aaa/ChangeUserRoles.java
```diff
@@ -0,0 +1,112 @@
...
+    public String getAPIPath() {
+        return "/aaa/changeRoles";
+    }
+
+
+    @Override
+    public JSONObject serviceImpl(Query query, HttpServletResponse response, Authorization authorization, final JsonObjectWithDefault permissions) throws APIException {
+
+        JSONObject result = new JSONObject();
+        if (authorization.getBaseUserRole().toString().equals("ADMIN")) {
```
Fluepke: @Orbiter Is this check necessary? As I understood you, it's enough to override `getMinimalBaseUserRole()`.

src/ai/susi/server/api/aaa/ChangeUserRoles.java
```diff
@@ -0,0 +1,112 @@
...
+                    serviceList.put(service.getCanonicalName());
+                }
+                result.put("serviceList", serviceList);
+                return result;
+            } else if (query.get("getUserRolePermission", false)) {
+                result.put("userRolePermissions", authorization.getUserRole().getPermissionOverrides());
+                return result;
+            } else {
+                // Accept POST params
+                String userTobeUpgraded = query.get("user", null);
```
Fluepke: Is it possible that an admin (the only admin) downgrades himself to a normal user resulting in a situation without any admin?

src/ai/susi/server/api/aaa/ChangeUserRoles.java
dynamitechetan: yes. I tried to set MinimalBaseUserRole as ADMIN, but it didn't worked. So I just wrote another condition to check it.

src/ai/susi/server/api/aaa/ChangeUserRoles.java
dynamitechetan: Yes. Its possible now but why will he do so. 
Also the first Admin has to make himself the admin by changing the value of userRole in the db/authorization.json.
For example, first orbiter can be the admin by changing the value of his userRole in authorization.json and then he can give rights to other admins.

src/ai/susi/server/api/aaa/ChangeUserRoles.java
```diff
@@ -0,0 +1,112 @@
...
+                String upgradedRole = query.get("role", "user");
+                // Validation
+                if(userTobeUpgraded==null){
+                    throw new APIException(400, "Bad username");
+                }
+                if(!DAO.authorization.has(userTobeUpgraded)){
+                    throw new APIException(400, "Username not found");
+                }
+                if(!DAO.userRoles.has(upgradedRole))
```
dynamitechetan: Okay.

Orbiter:
I moved the access right check to AbstractAPIHandler in https://github.com/fossasia/susi_server/commit/ed85d9690f901590a6189b02292293a2b453f1a3

There should never be the need to check the user rights inside the servlet, thats the purpose of the generic authorization.

If you write 'it is not working', then please:
- write down what you expected
- write down how you tested
- write down what was the result

So we can discuss about that.

Issue #289, opened 2017-06-14
Add Schema for Handling Groups in AAA


dynamitechetan:
Currently, there is no schema for handling groups. 
I will add a JSON Schema and a set of blank files so that we can work on them collaboratively.

Orbiter:
What is the purpose of groups? This is not a criticism to do that, in fact I believe that we do need the groups but how do we handle them in the context of authorization?

I believe a group should define which servlets can be accessed by a user which is otherwise not allowed to use the servlet. Therefore a group should have a list of servlets that are granted access.

Is that the concept with groups?

dynamitechetan:
Oh. Okay. I thought like we can have user groups like Linux. 
For example create a group for admins and a separate groups for privileged users. 
Different groups of a privileged users can have different permissions and access rights. 
Something like that. But as you are suggesting,  we can also have groups for users which limits their access to some servlets.

PR #290, opened 2017-06-14
Added JSON Schema for handling groups.

dynamitechetan:
Fixes issue #289 
I've added a JSON Schema for handling groups and blank java files to create API Endpoints. 
@saurabhjn76 and I will open issues separately for each API Endpoint.


dynamitechetan:
@saurabhjn76 @DravitLochan  please review

dynamitechetan:
I thought we can have user groups like Linux.
For example, create a group for admins and separate groups for privileged users.
Different groups of privileged users can have different permissions and access rights.

Issue #292, opened 2017-06-14
Add API Endpoints to List Details of a group


dynamitechetan:
I will be adding API Endpoints to list the details of a group in AAA

PR #296, opened 2017-06-16
API Endpoint to get details of a group

dynamitechetan:
Fixes issue #292 
This servlet adds an API Endpoint to return the details of the group whose name is passed in the URL GET parameter
This can be tested on http://127.0.0.1:4000/cms/getGroupDetails.json?group={groupName}


dynamitechetan:
okay

dynamitechetan:
@Orbiter, removed Groups.java and used JSONTray for storage.

PR #317, opened 2017-06-26
Removed groups.json, Fixes #316

### LINKED ISSUE #316: File 'groups.json' in data/settings should not be tracked by git
**Actual Behaviour**

in commit https://github.com/fossasia/susi_server/commit/45cdd0d8a34021557ece64bf8e6993eff5b33069 the file groups.json was added to git within the path data/settings.
The data path should be inside of gitignore and not be tracked by git.
Files cannot be added there.

**Expected Behaviour**

If susi_server expects that file upon startup, it must be generated.

**to do**

- remove the file, revert https://github.com/fossasia/susi_server/commit/45cdd0d8a34021557ece64bf8e6993eff5b33069
- add the file to conf/groups.json
- upon start check if the file exist in data/settings and if not create it by using a copy from conf/groups.json

dynamitechetan:
Fixes issue #316 
- [x]     removed the file from data folder
Deleted Groups.json and added groups.json to Gitignore

- [x]     add the file to conf/groups.json
_I thought that is not required. If the file does not exist a file is created automatically._

- [x]     upon start check if the file exist in data/settings and if not create it by using a copy from conf/groups.json
_Again if the file does not exist a file a created with an empty JSON Object. I tested it with getAllGroups and CreateGroups and it works fine._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ dist/loklak.jar
 
 /docs/_build
 
+data/settings/groups.json
--- a/data/settings/groups.json
+++ b/data/settings/groups.json
@@ -1,4 +1,0 @@
-{"group:group_Name": {
-  "group_members": {},
-  "groupBaseRole": "admin"
-}}


Orbiter:
Thank you!

Issue #261, opened 2017-06-07
Memory servelet support for displaying entire chat history


uday96:
**Actual Behaviour**

As of now, the clients use the memory servelet - https://github.com/fossasia/susi_server/commit/7d6245904c995a3aa5452e1ddbc99c35bddc9d86 to get chat history. But the servelet returns only 10 recent messages.

**Expected Behaviour**

We need a method in which we must be able to get the entire user history and not just a recent 10 or few.

Few ideas:
 1.  Removing the cognitions limit in the servelet and return the entire history. (This is a bad approach I guess cause the history could be huge?)
 2.  A method to give history messages as a group of a few (say 10) based on a request parameter determining how recent messages must be returned.
 For ex: suppose a user has 100 messages in total numbered 1 - 100 with 1 being most recent and every query to servelet returns 10 messages,  and say the request parameter is ```groupNo```
   So request like http://api.susi.ai/susi/memory.json?groupNo=0 must return messages 1-10
   and http://api.susi.ai/susi/memory.json?groupNo=1 returns 11-20 and so on, handling corner cases.
Something like this would allow the client to dynamically query history as the user scrolls up.

Please do suggest any other alternatives. The basic necceessity is to be able to access the entire chat history of the user

**UPDATE:**
A working solution as of now:
  1. Client sends the last message's identifier (preferably ```query_date``` or ```answer_date``` as of now or a new ```message id``` could be introduced) as a request parameter
  2. Server searches for the particular message using the identifier from the request param, in the cognitions array and returns the next few messages(say min(10,messages left)) along with the count of messages left in the user history (so that client knows when to stop querying for history)
This way client can dynamically query history accordingly without worrying about sync.

chiragw15:
@uday96 in second approach we will also need to send number of messages along with reponse. Like suppose total number of messages of user are 27. So, first and second response will include "numberOfMessages = 10" , the third one will include "numberOfMessages = 7" and the fourth one will include "numberOfMessages = 0". So, now since we have numberOfMessages = 0 , we know that there are no more messages, so we will stop sending a request for more results. What say?

saurabhjn76:
@uday96 @chiragw15 sounds good !!

uday96:
[...]

PR #323, opened 2017-06-27
feat : Improve memory servlet to return complete history

chiragw15:
Fixes issue #261 

Changes: Changed memory servlet to return complete history max 10 at a time.
new call will be 
http://127.0.0.1:4000/susi/memory.json?cognitions=5&date=2017-06-27T18:6:12.389Z
User will get next 5 cognitions starting from date given. Also added "cognitions_remaining" to indicate when user must stop querying. 
Use case : 
Suppose a client calls 
http://127.0.0.1:4000/susi/memory.json?cognitions=5
he will get first 5 cognitions, then after that he can call 
http://127.0.0.1:4000/susi/memory.json?cognitions=5&date=DATE_OF_FIFTH_COGNITION
this will give him next 5 cognitions, after that he can call
http://127.0.0.1:4000/susi/memory.json?cognitions=5&date=DATE_OF_TENTH_COGNITION
and so on.


chiragw15:
@dynamitechetan @saurabhjn76 @DravitLochan @uday96 review please.

src/ai/susi/server/api/susi/UserService.java
```diff
@@ -19,8 +19,9 @@
 
 package ai.susi.server.api.susi;
 
-import java.util.List;
+import java.util.*;
```
Orbiter: don't do that. There is no advantage. This is a no-sense-refactoring. I like to have the exact list of all classes which are used.

chiragw15:
@Orbiter done.

Issue #327, opened 2017-06-29
implement new action type : set alarm


DravitLochan:
**Expected Behaviour**
implement an action type which allows users to set alarms.

**Steps to reproduce it**
handle query at server and respond accordingly to clients.

**Would you like to work on the issue?**
Yes

DravitLochan:
Initially what we can do is that when the server receives a request, it is processes it and sends the response to the respective client making the request.
@Orbiter Are we also handling the case that when a user makes a call for alarm via web chat client, it also sets his alarm in android and iOS?

DravitLochan:
Should it be implemented as rss type or other action types (where users can make skills using that action type)?

Orbiter:
I dont't know how to explain. We discussed this in detail. If you believe a rss action type is an option here then you did not understand why we use actions and specific actions for the one and only purpose of that action only.

Hint: https://github.com/fossasia/susi_server/blob/development/src/ai/susi/mind/SusiAction.java#L49

Orbiter:
this kind of actions become superfluous with timed actions, see https://github.com/fossasia/susi_linux/issues/521

PR #334, opened 2017-06-30
New action type : alarm_set

DravitLochan:
First step #327  (not complete yet)

Changes: added new action type.


DravitLochan:
@Orbiter Please check that if I am doing it right. 
Thank you

src/ai/susi/mind/SusiSkill.java
```diff
@@ -145,6 +145,18 @@ else if (bang_type.equals("console")) {
                                             boa.has("latitude") && boa.has("longitude") && boa.has("zoom")) {
                                         actions.put(SusiAction.mapAction(
                                             boa.getDouble("latitude"), boa.getDouble("longitude"), boa.getInt("zoom")));
+                                    } else
+                                    if(type.equals(SusiAction.RenderType.alarm_set.toString()) &&
+                                            boa.has("hour")){
+                                        int hour = boa.getInt("hour");
+                                        if(hour>12)
+                                            actions.put(SusiAction.alarmSetAction(
+                                                    hour-12, boa.has("minute") ? boa.getInt("minute") : 0,
```
Orbiter: I don't think that hour-12 is a good idea, do you think that this should actually work? How do you set am or pm?

src/ai/susi/mind/SusiAction.java
```diff
@@ -46,6 +46,7 @@
         websearch,     // do a web search on the client, show like rss rendering
         anchor,        // show/say a link
         map,           // show a map
+        alarm_set,     //set alarm
```
Orbiter: no, that is done with time_set. How is alarm_set different from timer_set?

DravitLochan: Those are 2 different options that mobile phones offer. So I thought that it would be good to handle casses. Should I still change to timer?

src/ai/susi/mind/SusiSkill.java
DravitLochan: This is to handle according to a 24 hour clock.

Orbiter:
What we need here is also a common view what the hour, minute second expresses: a delta to a time or an absolute time? I would consider this an absolute time. That means that all skills which use a relative time in an intent must do a translation of a relative time to an absolute time.

Issue #345, opened 2017-07-06
separate implementation of cahnge password and password recovery


DravitLochan:
**Actual Behaviour**
currently if a user has to reset/change his password they need to go to recover password. From there a mail is sent to their email id which has a token in it. This link allows user to reset the password. 

**Expected Behaviour**
password recovery and reset/change password are 2 different things. If a user knows his current password, simply allow them to change their password.

**Steps to reproduce it**
Add a new servlet which takes current password and new password of the user logged in. If user is not logged in, ask them to login first.

**Would you like to work on the issue?**
yes

PR #352, opened 2017-07-07
	API endpoint to allow user to change password

DravitLochan:
Fixes issue #345 
Changes: added API end point for change password
test locally at http://127.0.0.1:4000/aaa/changepassword.json
send the following three parameters :
changepassword : contains email id of the user who's password has to be changed
password : current password
newpassword : new password

### DIFF
--- a/src/ai/susi/server/api/aaa/PasswordChangeService.java
+++ b/src/ai/susi/server/api/aaa/PasswordChangeService.java
@@ -31,6 +31,11 @@ import java.util.regex.Pattern;
 
 /**
  * Created by dravit on 6/7/17.
+ * test locally at http://127.0.0.1:4000/aaa/changepassword.json
+ * send the following three parameters :
+ * changepassword : contains email id of the user who's password has to be changed
+ * password : current password
+ * newpassword : new password
  */
 public class PasswordChangeService extends AbstractAPIHandler implements APIHandler{
     @Override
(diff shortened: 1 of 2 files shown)


DravitLochan:
@dynamitechetan @saurabhjn76 @madhavrathi @jyoth1raditya Please reivew!

DravitLochan:
@dynamitechetan @saurabhjn76 @jyoth1raditya @Orbiter Please review!

src/ai/susi/server/api/aaa/PasswordChangeService.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public ServiceResponse serviceImpl(Query post, HttpServletResponse response, Authorization rights, JsonObjectWithDefault permissions) throws APIException {
+        JSONObject result = new JSONObject();
+        result.put("accepted", false);
+
+        String useremail = post.get("changepassword", null);
+        String password = post.get("password", null);
+        String newpassword = post.get("newpassword",null);
```
sudheesh001: Sending passwords to the server as plaintext doesn't seem to be the right thing to do. Can you send a hash and then decrypt the hash on the server if needed, ideally that shouldn't be done ever. Storing hashes and checking for hashes seems to be more promising.

But I think that is however a much larger pull request than the one being looked at here.

DravitLochan: Yeah.! I too believe that sending password as plain text is a bit compromising with security of users, but same policy is followed in login, signup and reset password. So I believe if we have to change it, it needs to be done across all end points.

dynamitechetan: Sending plain text passwords over https is quite secure, the traffic is encrypted. Of course, storing the passwords needs proper encryption,

Issue #377, opened 2017-07-15
Handle Change password in server


madhavrathi:
**Actual Behaviour**

Currently server is not handling the case where user changes his password to current one.

**Expected Behaviour**

Server should send a prompt which will be shown in respective clients that "Please enter a password different from current one"

**Would you like to work on the issue?**

I will try to handle it but if not successful will ask developers from server team to look into this.
Thanks

PR #378, opened 2017-07-15
Fixes#377 Same password message added in PasswordChangeService.java

madhavrathi:
Fixes issue #377 

Changes: Now if the current password and new password are same, It is handled in server and message prompts "Your current password and new password matches".

Screenshots for the change: 
issuehbhaika
@dynamitechetan @DravitLochan @mariobehling @Orbiter Please Review
Thanks

### DIFF
--- a/src/ai/susi/server/api/aaa/PasswordChangeService.java
+++ b/src/ai/susi/server/api/aaa/PasswordChangeService.java
@@ -62,55 +62,62 @@ public class PasswordChangeService extends AbstractAPIHandler implements APIHand
         String password = post.get("password", null);
         String newpassword = post.get("newpassword",null);
 
-
-        ClientCredential pwcredential = new ClientCredential(ClientCredential.Type.passwd_login, useremail);
-        Authentication authentication = DAO.getAuthentication(pwcredential);
-        ClientCredential emailcred = new ClientCredential(ClientCredential.Type.passwd_login,
-                authentication.getIdentity().getName());
-        ClientIdentity identity = authentication.getIdentity();
-        String passwordHash;
-        String salt;
-
-        try {
-            passwordHash = authentication.getString("passwordHash");
-            salt = authentication.getString("salt");
-        } catch (Throwable e) {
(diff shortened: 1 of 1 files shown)


src/ai/susi/server/api/aaa/PasswordChangeService.java
```diff
@@ -62,55 +62,62 @@ public ServiceResponse serviceImpl(Query post, HttpServletResponse response, Aut
...
-
-        try {
-            passwordHash = authentication.getString("passwordHash");
-            salt = authentication.getString("salt");
-        } catch (Throwable e) {
-            Log.getLog().info("Invalid password try for user: " + identity.getName() + " from host: " + post.getClientHost() + " : password or salt missing in database");
-            result.put("message", "invalid credentials");
-            throw new APIException(422, "Invalid credentials");
+        if(password.equals(newpassword)){
+            result.put("message", "Your current password and new password matches");
```
DravitLochan: give an appropriate message. this message does not helps the user to understand that did the password finally change or not.

madhavrathi: Sure will update the message. Please suggest an appropriate message. Thanks

madhavrathi: Also we have to follow this approach as we want the message to be displayed, as user setting his new password similar to current one is not an error. @dynamitechetan Don't you think this is the approach we should. Please give your inputs. Thanks

DravitLochan: This is not the best approach because an error should always be sent as an error. But since catching and showing error in ReactJS is not possible, This is the next best approach. So ya.. go on with this for now. Also, the error message could be more informative like "Current password and New password can not be similar".

madhavrathi: Agreed, also as this message will be required in all clients so this is the way we should handle it.

PR #389, opened 2017-07-19
added !description to the skill for description

### LINKED ISSUE #383: Add description and images to skills
**Actual Behaviour**
Currently, the skills lack description, and image needed to create a skill page at skill_susi_cms like that of Alexa 

image

**Expected Behaviour**
!description and !images tags should be added to rules to fetch description and images URL from skills.

Yes, I will be looking  !example implementation for reference.

saurabhjn76:
progress on #383 

Changes: added new !description to skill

### DIFF
--- a/src/ai/susi/mind/SusiIntent.java
+++ b/src/ai/susi/mind/SusiIntent.java
@@ -59,7 +59,7 @@ public class SusiIntent {
     private final int user_subscore;
     private Score score;
     private final int id;
-    private final String skill, example, expect;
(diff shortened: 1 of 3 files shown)


saurabhjn76:
@dynamitechetan @DravitLochan @madhavrathi @jyoth1raditya please review

mariobehling:
We need to think more in detail how we are going to handle the images. They should be stored in the system in a standard location. Please follow up separately.

Issue #417, opened 2017-07-27
remove redundant link from in case of map action


chashmeetsingh:
**Actual Behaviour**

Both the anchor action and answer action contain the same link.

**Expected Behaviour**

Only the anchor action should contain the link for the map.

**Steps to reproduce it**

Query: `where is delhi`.

**Would you like to work on the issue?**

No.

yashLadha:
Need to close this issue @mariobehling

PR #418, opened 2017-07-27
Removed Open street maps link form answer causing problems in TTS

dynamitechetan:
Fixes issue #417 
Removed Open street maps link form answer causing problems in TTS.


mariobehling:
Should this be removed on the server or taken care by the client?

dynamitechetan:
Clients have to remove the link from the answer manually. Also, there is a link present in the anchor object already, so if we want to show a map we can use the anchor object.
For replying using TTS, the answer containing text should be enough.

Issue #427, opened 2017-07-31
Create new file on modifyskill.json if skill does not exists


dynamitechetan:

mariobehling:
What is this issue about? Why is there no description? What is your workflow? Quickly open an issue and add code that you have done before? This does not make sense.

dynamitechetan:
Sorry, I forgot to open the issue previously. It was a small change, so I made the issue just before opening a PR.
Sorry I will not repeat this.

PR #428, opened 2017-07-31
Create new file on modify skill if skill does not exist

dynamitechetan:
Create a new file on modify skill if skill does not exist.
Fixes issue #427


mariobehling:
Please also document this and stick to the proper workflow and best practices!

Issue #433, opened 2017-08-01
Change Modify Skill Service base role to User from anonymous


dynamitechetan:
Currently, Modify Skill Service has a base role of anonymous. We need to change it to User.

PR #435, opened 2017-08-01
Changed BaseRole in Create & modify skill service from Anonymous to USER

dynamitechetan:
Changed Base Role in Create & modify skill service from Anonymous to USER.
Currently, any Anonymous user can create and modify a skill.
Fixes issue #433


dynamitechetan:
as per conversation last night, we need skills to be edited by ADMINS only. So I update the PR and changed the base role as ADMIN.

dynamitechetan:
Yes. Changing it right away.

madhavrathi:
@dynamitechetan yes please change it accordingly.

dynamitechetan:
Done!

PR #438, opened 2017-08-01
Remove Travis deploy docs and update subprojects (#426)

imujjwal96:
* Remove Travis deploy docs and update subprojects

* Code cleanup

Fixes issue #[Add issue number here. If you do not solve the issue entirely, please change the message e.g. "First steps for issues #IssueNumber]

Screenshots for the change:


.travis.yml
```diff
@@ -20,26 +20,12 @@ after_success:
 - bash <(curl -s https://codecov.io/bash)
 - bash kubernetes/travis/deploy.sh
 - codecov
-before_deploy:
```
pri22296: You sure It's not needed anywhere else?

imujjwal96: The key is required to clone and publish docs using ssh. I dont think it'll be needed anywhere else.

imujjwal96:
@pri22296 Also, build was successful 
image

Issue #445, opened 2017-08-04
First user who sign's up will become admin


DravitLochan:
**Actual Behaviour**
Currently, Those who want to become SysAdmins should have access to server files.They need to manually change user role in authorization.json file.

**Expected Behaviour**
After installing SusiServer, user should be prompted to signup as SysAdmin. ( First user to become SysAdmin).

**Would you like to work on the issue?**
Yes

Issue #446, opened 2017-08-04
remove "host:ip address" from username array returned from /aaa/getAllUsers.json 


DravitLochan:
**Actual Behaviour**
host:17.0.0.1 is added in the list of username arrays

**Expected Behaviour**
it should not be included.

**Would you like to work on the issue?**
Yes

PR #478, opened 2017-08-11
Remove host from getAllUser, first user to signup becomes admin and fix for deletion of account

DravitLochan:
Fixes issue #445 , #446 , #481 

Changes: First user to signup on the server will be made admin by default; while iterating through the list of authorized clients, if host is encountered then don't include it in the array; on deleting account, user removed from authorization file as well.
before : 
before

after : 
after

### DIFF
--- a/src/ai/susi/DAO.java
+++ b/src/ai/susi/DAO.java
@@ -358,6 +358,10 @@ public class DAO {
 		}
 		return i;
 	}
+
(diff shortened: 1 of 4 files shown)


DravitLochan:
@saurabhjn76 @dynamitechetan Please review. @mariobehling Please do not squash before merge as this pr is solving 2 issues simultaneously.

DravitLochan:
@Orbiter Yeah! Maximum rights now

PR #495, opened 2017-08-13
Fixes #226 changes for AutomaticDowload of public-transport-enabler

### LINKED ISSUE #226: replace public transit submodule with downloaded library
As discussed with @mariobehling we want to keep it as premise to not use submodules in this project. Submodules will cause deployment problems.

The public transit library as introduced in https://github.com/fossasia/susi_server/pull/191 shall be replaced with a direct download command in the gradle build script. The build script must check if the library has been downloaded so far and do nothing in that case.

code stub hint: should be something like 
```
def f = new File('the file path')
if (!f.exists()) {
    new URL('the url').withInputStream{ i -> f.withOutputStream{ it << i }}
}
```

fazeem84:
Fixes #226 changes for AutomaticDowload of public-transport-enabler  

Gradle changes for  of public-transport-enabler AutoDownload
Used Gradle-git plugin for  GitCloning functionality


fazeem84:
Could you please solve coverage issue?

fazeem84:
@Orbiter  could you please help.me to fix code coverage issue?

dynamitechetan:
I think codecov is not setup properly. For now please ignore that. Just make sure travis passes.

Issue #497, opened 2017-08-13
Send user role in getUsers end point


DravitLochan:
From aaa/getUsers.json?access_token=go2ijgk5ijkmViAac2bifng3uthdZ&page=2 following sample data is returned : 
{
  "accepted": true,
  "users": [
    {
      "type": "email",
      "name": "hi@susi.com",
      "anonymous": false,
    },
    {
      "type": "email",
      "name": "hi@test.com",
      "anonymous": false,
    },
    {
      "type": "email",
      "name": "hi@fossasia.com",
      "anonymous": false,
    }
  ],
  "username": [
    "email:hi@susi.com",
    "email:hi@test.com",
    "email:hi@fossasia.com"
  ],
  "message": "Success: Fetched all Users!",
  "session": {"identity": {
    "type": "email",
    "name": "hi@susi.com",
    "anonymous": false
  }}
}

data returned should also contain user role in it. New data type would be 
{
  "accepted": true,
  "users": [
    {
      "type": "email",
      "name": "hi@susi.com",
      "anonymous": false,
      "userRole": "ADMIN"
    },
    {
      "type": "email",
      "name": "hi@test.com",
      "anonymous": false,
      "userRole": "USER"
    },
    {
      "type": "email",
      "name": "hi@fossasia.com",
      "anonymous": false,
      "userRole": "USER"
    }
  ],
  "username": [
    "email:hi@susi.com",
    "email:hi@test.com",
    "email:hi@fossasia.com"
  ],
  "message": "Success: Fetched all Users!",
  "session": {"identity": {
    "type": "email",
    "name": "hi@susi.com",
    "anonymous": false
  }}
}

edits : 
In my pr, I have used lower case for user roles.

DravitLochan:
fixed!

PR #498, opened 2017-08-13
Added UserRole in user array for each user

DravitLochan:
Fixes issue #497 
Minor bug fix : If a page is requested which does not exists, throw error to user with message list does not exists.


DravitLochan:
@Orbiter @saurabhjn76 @dynamitechetan Please review!

Issue #496, opened 2017-08-13
update ChangeUsers end point


DravitLochan:
**Actual Behaviour**
Server responds with 500 error code when called http://api.susi.ai/aaa/changeRoles?user=lochandravit@gmail.com&role=admin called. 

**Expected Behaviour**
1. When called, user role of user (i.e. lochandravit@gmail.com)  should change.
2. Servlet should be able to respond with all the possiblr user roles when a client requests them.

**Would you like to work on the issue?**
yes
screenshot : 
![error](https://user-images.githubusercontent.com/14837478/29249002-c7d7d344-8043-11e7-85ee-fdccc7b813e5.png)

PR #504, opened 2017-08-14
ChnageUserRole bug fix

DravitLochan:
Fixes issue #496 
Changes: 
1. added .json in the API end point.
2. change user role was not working as client identity was throwing error while generating one for the user whose role was to change.

Screenshots for the change: 
image

### DIFF
--- a/src/ai/susi/server/api/aaa/ChangeUserRoles.java
+++ b/src/ai/susi/server/api/aaa/ChangeUserRoles.java
@@ -14,8 +14,11 @@ import java.lang.reflect.Constructor;
 /**
  * Created by chetankaushik on 30/05/17.
  * This is a servlet that is used to change the role of any user
- * This requires Admin login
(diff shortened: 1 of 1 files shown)


DravitLochan:
@Orbiter @mariobehling @dynamitechetan @saurabhjn76 Please review!

src/ai/susi/server/api/aaa/ChangeUserRoles.java
```diff
@@ -90,30 +93,61 @@ public ServiceResponse serviceImpl(Query query, HttpServletResponse response, Au
             result.put("message", "Successfully processed request");
             return new ServiceResponse(result);
         } else {
-            // Accept POST params
+            // Accept POST/GET params
```
dynamitechetan: This actually works on both GET and POST. Rest looks good.

DravitLochan: Yep. That is the reason I updated the comments for get/post.

Issue #487, opened 2017-08-11
Keep track of last login ip, last login time and signup time


DravitLochan:
@Orbiter Please tell in which file should these details be stored? 
accounting, authorization or authentication. Or should a new file be added for such details?

Orbiter:
Within the AAA concept this belongs to Accounting: "what has the user done".
No new file. Add a json object for a users accounting object and write that there.

PR #505, opened 2017-08-14
return accounting attributes in response of getUser

DravitLochan:
Fixes issue #487 
Changes: after this pr, server will store last login IP of the user and will return last login IP, last login time and signup time. If any of the details is missing, it will return an empty string.
Server will also tell if user is verified or not.

Screenshots for the change: 
image


DravitLochan:
@Orbiter @saurabhjn76 @dynamitechetan @madhavrathi Please review.

PR #520, opened 2017-08-16
Feat: Delete Skill Stores a deleted skill for 30 days

### LINKED ISSUE #486: Update Delete Skill Endpoint
1. Delete Skill Endpoint should also be able to delete a skill. The skill should be moved to an area "deleted skills" and be deleted automatically after 30 days.
2. Admins should have an option to re-add the skill within that period as well.

dynamitechetan:
Fixes #486  
Delete Skill Endpoint is now able to delete a skill. 
The skill should be moved to an area "deleted skills" and be deleted automatically after 30 days.
http://127.0.0.1:4000/cms/deleteSkill.txt?skill=gender

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ def versionJersey = '2.22.2'
 buildscript {
   repositories {
     jcenter()
+    mavenCentral()
   }
 
   dependencies {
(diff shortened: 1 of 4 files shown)


mariobehling:
@dynamitechetan Please explain where are the skills stored in the meantime for the 30 days in the repository? Are you adding a "delete" tag or move them to a dedicated "delete" folder? Please add the relevant info to the documentation.

dynamitechetan:
@mariobehling When someone deletes a skill then it will move a folder ```delete_skills_dir```.
When a file is moved to the ```delete_skills_dir``` its last modified date is changed to the current date.
Then in the caretaker there is a function which checks for files which are older than 30 days by checking the last modified date.
If there is any file which is older than 30 days then it deletes them.

mariobehling:
Ok, got it. Thanks!

mariobehling:
@dynamitechetan Add this info to the documentation in following PRs, please.

Issue #513, opened 2017-08-16
Bulky Docker image for Kubernetes


singhpratyush:
The Docker images for Kubernetes is too large. This is mainly because of using Ubuntu as base image and having too many layers while build.

### Steps to reproduce -

1. Build the Kubernetes Docker image
2. Take a look at the images size

AnshulMalik:
Hey, are you talking about [this](https://github.com/fossasia/susi_server/blob/development/Dockerfile) dockerfile?

singhpratyush:
@AnshulMalik: No, it is this one - https://github.com/fossasia/susi_server/blob/development/kubernetes/images/Dockerfile.

AnshulMalik:
Okay, I'll look into it :)

imujjwal96:
the ubuntu image itself is a lot bloated. Try using other images.

singhpratyush:
@AnshulMalik: This may be helpful - http://blog.fossasia.org/optimising-docker-images-for-loklak-server/.

AnshulMalik:
That's good article

abishekvashok:
@singhpratyush pls review #582

PR #582, opened 2017-10-01
Reduce size for Kubernetes docker image.

davgomgar:
The size of the Kubernetes image was reduce by using alpine as base image instead of Ubuntu and by reducing the number of layers

First steps for #513

Changes: 
- The base image was changed from Ubuntu to Alpine Linux.
- The number of layers were reduced by grouping RUN commands together
- Change MAINTAINER (deprecated) for a LABEL as suggested in Docker documentation (https://docs.docker.com/engine/reference/builder/) 

Maybe there is room for more improvement in this, but at the moment I'm learning Docker and I'm not sure if I will be able to improve it much more :)

### DIFF
--- a/kubernetes/images/Dockerfile
+++ b/kubernetes/images/Dockerfile
@@ -1,46 +1,30 @@
-FROM ubuntu:latest
-MAINTAINER Ansgar Schmidt <ansgar.schmidt@gmx.net>
-ENV DEBIAN_FRONTEND noninteractive
+FROM alpine:latest
+LABEL maintainer="Ansgar Schmidt <ansgar.schmidt@gmx.net>"
 
 ARG BRANCH
 ARG COMMIT_HASH
 
 ENV COMMIT_HASH ${COMMIT_HASH:-null}
 ENV BRANCH ${BRANCH:-development}
-
-# update
-RUN apt-get update
-RUN apt-get upgrade -y
-
-# add packages
-RUN apt-get install -y git openjdk-8-jdk
-
-RUN cd / && git clone https://github.com/fossasia/susi_skill_data.git
-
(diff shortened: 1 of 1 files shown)


royjesslong:
Reduce

kubernetes/images/Dockerfile
```diff
@@ -1,46 +1,30 @@
...
-
-# hack until susi support no-daemon
-RUN echo "while true; do sleep 10;done" >> bin/start.sh
+RUN git checkout ${BRANCH} && \
+    if [ -v COMMIT_HASH ] ; then git reset --hard ${COMMIT_HASH} ; fi && \
+    git submodule update --recursive --remote && \
+    git submodule update --init --recursive
+
+# remove git history to reduce size of image, compile, change config file and hack until susi support no-daemon
+RUN rm -rf dependencies/public-transport-enabler/.git && \
```
singhpratyush: This can be combined with the above `RUN` commands.

kubernetes/images/Dockerfile
```diff
@@ -1,46 +1,30 @@
...
+    rm -rf .git && \
+    ./gradlew build -x test && \
+    sed -i.bak 's/^\(port.http=\).*/\180/'        conf/config.properties && \
+    sed -i.bak 's/^\(port.https=\).*/\1443/'      conf/config.properties && \
+    sed -i.bak 's/^\(upgradeInterval=\).*/\186400000000/' conf/config.properties && \
+  echo "while true; do sleep 10;done" >> bin/start.sh
```
singhpratyush: Indentation issue here.

sudheesh001:
What's the new image size with alpine instead of ubuntu?

PR #250, opened 2017-06-05
Added API endpoint for listing models 

saurabhjn76:
Fixes issue #248

Changes: Added ModelListService for adding API endpoint.
Screenshots for the change: 
image
This can be tested at : [http://127.0.0.1:4000/cms/getmodel.json](http://127.0.0.1:4000/cms/getmodel.json)

PR #583, opened 2017-10-01
deleted superfluous UTF8 class and bad usernames

bryanlou:
Fixes #250 

Changes: SusiCognition in server directory


mariobehling:
Thanks. Please re-commit. As I understand it is messed up somehow.

PR #601, opened 2017-10-20
remove redundant mavenCentral repo

garg-anuj:
jcenter() is superset of mavenCentral() so it is redundant to use mavenCentral after jcenter.


garg-anuj:
Lets merge it now

Issue #618, opened 2017-12-10
Unable to build with Java 9


yashLadha:
**Actual Behaviour**

Error while building the setup.

**Expected Behaviour**
Build should be successfull

**Steps to reproduce it**

Install with java 9 environment.

**LogCat for the issue**

Could not determine java version from '9.0.1'.

**Screenshots of the issue**

![image](https://user-images.githubusercontent.com/18033231/33802832-c607ef48-dda6-11e7-892c-cbda955f17fc.png)

abishekvashok:
@yashLadha update grade to the latest rc version.

yashLadha:
But i executed the script that is mentioned in the documentation. May be changes needs to be done in that @abishekvashok

abishekvashok:
@yashLadha try updaitng the gradle version in the repo or degrading your java

yashLadha:
I have degraded the java version but the tests are failing so it is not able to build @abishekvashok :sweat_smile:

Orbiter:
not a bug - we are using java 8

PR #619, opened 2017-12-10
Trying to fix 618

niteshkumarniranjan:
Fixing #618

Changes: .travis.yml


Orbiter:
we are not using java 9 features so any fix going into that direction is misleading

PR #622, opened 2017-12-13
Stop preload.sh from killing service

### LINKED ISSUE #621: Running start.sh twice kills the service
**Actual Behaviour**

running start.sh twice kills the service 

**Expected Behaviour**

If the service is already running, it should everytime say  --

`Server is already running, please stop it and then start`
if the service hasn't stopped yet.

**Steps to reproduce it**

Run the start.sh twice and the service will be stopped.

Yes

harsh-98:
Fixes #621


marvin08:
@harsh-98 Please refer to my comment on #621.

PR #652, opened 2018-01-12
Fixes #651: Moves environmental variables to respective block

### LINKED ISSUE #651: Move environmental variables in travis.yml to repsective block
**Actual Behaviour**

Codecov is tested on another vm instance. It does not require secure variables.

**Expected Behaviour**
It should be opresent in the instance that does deployment only.

**Steps to reproduce it**

See the travis ci log

Yes!.

abishekvashok:
Moves environmental variables to be present only in the instance that
does deployment as they are only used for deployment.

Fixes #651

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,18 @@
 dist: trusty
 sudo: required
 
-# Environment Variables
-env:
-  global:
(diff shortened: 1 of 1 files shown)


abishekvashok:
Thanks @harshit98

abishekvashok:
@fossasia/susi-ai-admins can we have a merge?

abishekvashok:
Ping @daminisatya

abishekvashok:
Thanks @the-dagger

Issue #672, opened 2018-03-04
Live query response


AnupKumarPanwar:
**Idea**
I would like to add a feature of live query response below the textbox, so that it is convenient for the developer to test the response of the query before looking into the actual JSON response. 

**Mock**
![image](https://user-images.githubusercontent.com/10573038/36944474-f8f27852-1fc2-11e8-993f-12ee9ce277d4.png)

PR #675, opened 2018-03-04
Added live query response

AnupKumarPanwar:
Fixes [Live query response #672]

Changes
Added live query search on the index page using jQuery. It shows the SUSI reply as soon as the user types in the input field. User can click the "Get JSON Data" button or press Enter to get the full JSON response from the server.

Screenshots for the change: 

screenshot from 2018-03-04 15-37-49

### DIFF
--- a/html/index.html
+++ b/html/index.html
@@ -36,9 +36,37 @@
   }
 }
     </script>
+    <script src="./js/jquery.min.js"></script>
     <!--#include file="../ssi/common_head_back.include" -->
     <!--#include file="../data/ssi/custom_head.include" -->
   </head>
+
+  <script>
+    $(document).ready(function(){
+      var timezoneOffset=new Date().getTimezoneOffset();
+      var xhttp = new XMLHttpRequest();
+      $('#query').keyup(function()
+      {
+        xhttp.abort();
+        var q=$(this).val();
+
+        if (q.length==0) {
+          $("#liveResponse").text("");
+        }
+        else{
+              xhttp.onreadystatechange = function() {
+              if (this.readyState == 4 && this.status == 200) {
+               console.log(JSON.parse(this.responseText).answers[0].actions[0].expression);
+               $("#liveResponse").text(JSON.parse(this.responseText).answers[0].actions[0].expression);
+              }
+            };
+            xhttp.open("GET", "/susi/chat.json?timezoneOffset="+timezoneOffset+"&q="+q, true);
+            xhttp.send();
+        }
+
+      })
+    })
+  </script>
 
   <body style="position: relative; margin: 0; padding-bottom: 6rem; min-height: 100%;"> <!-- style woodoo is for footer line -->
     <!--#include file="../data/ssi/custom_body_front.include" -->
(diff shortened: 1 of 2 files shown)


AnupKumarPanwar:
@dynamitechetan @saurabhjn76 @Orbiter @DravitLochan Please review

yashLadha:
What if the expression is of some large lenght. @AnupKumarPanwar . Can you give some more ss to better understand the behaviour.

html/index.html
```diff
@@ -36,10 +36,38 @@
...
+               console.log(JSON.parse(this.responseText).answers[0].actions[0].expression);
+               $("#liveResponse").text(JSON.parse(this.responseText).answers[0].actions[0].expression);
+              }
+            };
+            xhttp.open("GET", "/susi/chat.json?timezoneOffset="+timezoneOffset+"&q="+q, true);
+            xhttp.send();
+        }
+
+      })
+    })
```
sudheesh001: When you've included `jquery` why are you still using vanilla js style of creating an XHR request explicitly? This can be done by `getJSON()` and with other related jQuery based Ajax implementations.

html/index.html
```diff
@@ -78,16 +106,22 @@ <h2>Scientific User Support Intelligence</h2>
             <input id="timezoneOffset0" type="hidden" name="timezoneOffset" value="">
             <script>document.getElementById("timezoneOffset0").value = new Date().getTimezoneOffset();</script>
             <div class="input-group input-group-lg col-sm-8" style="float:left">
-              <input name="q" id="query" type="text" class="center-block form-control input-lg" title="Search Twitter and the loklak network" placeholder="Chat with Susi" autofocus>
+              <input name="q" id="query" type="text" class="center-block form-control input-lg" title="Search Twitter and the Loklak network" placeholder="Chat with Susi" autofocus autocomplete="off">
```
sudheesh001: Any specific reason for this?

html/index.html
```diff
@@ -78,16 +106,22 @@ <h2>Scientific User Support Intelligence</h2>
...
+              <!--Show live query response here -->
+            </div>
+          </div>
+
           <form action="#" class="col-lg-12 hidden-lg hidden-md hidden-sm row">
             <input id="timezoneOffset1" type="hidden" name="timezoneOffset" value="">
             <script>document.getElementById("timezoneOffset1").value = new Date().getTimezoneOffset();</script>
             <div class="input-group input-group-lg row col-sm-12" style="float:left;display: flex;flex-wrap: wrap;">
-              <input name="q" id="query" type="text" class="center-block form-control input-lg" title="Search Twitter and the loklak network" placeholder="Search Twitter & the loklak network" style="border-bottom-width: 5px;padding-right: 12px;">
+              <input name="q" id="query" type="text" class="center-block form-control input-lg" title="Search Twitter and the Loklak network" placeholder="Search Twitter & the loklak network" style="border-bottom-width: 5px;padding-right: 12px;">
```
sudheesh001: Replace `&` as `&amp;`

html/index.html
AnupKumarPanwar: There is an onchange listener on the text field which creates a new request and deletes the previous one when the change occurs in the input field.

sudheesh001: Sure but I fail to understand why this is still written the way it is here. Why mix JS and jQuery? The onchange listeners and so on can also be done with jQuery and doesn't really need javascript in its vanilla version. Also, maybe please break the relevant js into a separate script file?

mariobehling:
I am not sure, if this is a good idea as it would increase the load on the server without a specific user requirement.

sudheesh001:
Closing for inactivity. Please reopen with updates. It'd be nice to have a toggle on the UI asking if the user wants live queries as they type ahead.

PR #728, opened 2018-05-19
Added an API to store information when a new device is connected

### LINKED ISSUE #727: Create an endpoint to store the device's (smart speaker) information in the User's accounting model
As discussed in the meeting, the list of connected devices are supposed to be displayed on the web client. Hence, there needs to be an API implemented which when called from the Android/iOS devices on SignUp using Wifi Direct, will store the information of the connected device (Smart speaker). The information will have to be stored in the accounting model of the particular User.

Yes

Akshat-Jain:
Fixes #727 

Changes: I've added a servlet whose endpoint, that is _/aaa/addNewDevice.json_, returns the information of the connected device (Smart Speaker) in a JSON format and also store that information in a file for future use. This stored information can then be retrieved and used to display the list of connected devices on the web client (as discussed in the meeting).

### DIFF
--- a/src/ai/susi/SusiServer.java
+++ b/src/ai/susi/SusiServer.java
@@ -472,6 +472,7 @@ public class SusiServer {
                 TopMenuService.class,
                 PasswordChangeService.class,
                 ListSettingsService.class,
+                AddNewDevice.class,
                 ConvertSkillJsonToTxtService.class,
                 GroupListService.class,
                 ConvertSkillTxtToJsonService.class,
--- a/src/ai/susi/server/api/aaa/AddNewDevice.java
+++ b/src/ai/susi/server/api/aaa/AddNewDevice.java
@@ -1,0 +1,131 @@
+/**
+ *  Copyright 19.05.2018 by Akshat Jain, @Akshat-Jain
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2.1 of the License, or (at your option) any later version.
+ *
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public License
+ *  along with this program in the file lgpl21.txt
+ *  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package ai.susi.server.api.aaa;
+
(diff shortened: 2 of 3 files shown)


Akshat-Jain:
@dynamitechetan @DravitLochan @saurabhjn76 Please review. Thanks :-)

Akshat-Jain:
@arundhati24 @batbrain7 @imjog Please review :-)

src/ai/susi/SusiServer.java
```diff
@@ -472,6 +472,7 @@ private static void setServerHandler(File dataFile){
                 TopMenuService.class,
                 PasswordChangeService.class,
                 ListSettingsService.class,
+								AddNewDevice.class,
```
PrP-11: Please check indentation here. :-)

Akshat-Jain:
@dynamitechetan I've made the requested changes. Please review :-)
Thanks :-)

src/ai/susi/server/api/aaa/AddNewDevice.java
```diff
@@ -0,0 +1,131 @@
...
+public class AddNewDevice extends AbstractAPIHandler implements APIHandler{
+
+    @Override
+    public String getAPIPath() {
+        return "/aaa/addNewDevice.json";
+    }
+
+    @Override
+    public UserRole getMinimalUserRole() {
+        return UserRole.ANONYMOUS;
```
dynamitechetan: This access level must be set to USER.

src/ai/susi/server/api/aaa/AddNewDevice.java
```diff
@@ -0,0 +1,131 @@
...
+            String s = result.toString();
+            byte b[]=s.getBytes();//converting string into byte array
+            fs.write(b);
+          }
+          catch(Exception e){
+            DAO.severe(e);
+          }
+
+          FileWriter fw = null;
+          BufferedWriter bw = null;
```
saurabhjn76: Move these declarations at the beginning of the method.

Akshat-Jain:
@dynamitechetan @saurabhjn76 Made the required changes :-)

Orbiter:
I see that here a new storage principle is introduced: a json file for each user ID. stored is the information email and an ID. I don't know why you want to store an email address using the user ID which is actually also an emails address (in most cases).

The whole API has no definition, which information shall be stored and in which data structure this shall happen. There is no abstraction layer of this information.

Furthermore, we already have a storage mechanism for user-related data: the accounting object. You should not store new information in a new data structure, instead create a class which handles your new data and adds it to the accounting object.

For an example on how to use the Accounting, see the class https://github.com/fossasia/susi_server/blob/development/src/ai/susi/server/api/aaa/ChangeUserSettings.java.

Please start from scratch: find out what information is needed to identify an device. Then make a data structure for this. Then investigate which API is already there and if you can use it.

PR #759, opened 2018-05-29
/cms/getSkillList.json endpoint returns list of all Skills if no query parameter is given

### LINKED ISSUE #705: getSkillList servlet needs modifications to show JSON data of all skills
**Actual Behaviour**

Currently, the getSkillList servlet is showing JSON data for only skills which fall under the group "Knowledge". 

**Expected Behaviour**

It should show JSON data of all skills.

**Steps to reproduce it**
Go to https://api.susi.ai/cms/getSkillList.json. We can see that JSON data is being shown for only the skills under the group "Knowledge".

Yes

Akshat-Jain:
Fixes #705 

Changes: Changes /cms/getSkillList.json endpoint to return list of all skills when no query parameter is given.

### DIFF
--- a/src/ai/susi/server/api/cms/ListSkillService.java
+++ b/src/ai/susi/server/api/cms/ListSkillService.java
@@ -43,7 +43,7 @@ public class ListSkillService extends AbstractAPIHandler implements APIHandler {
 
         String model_name = call.get("model", "general");
         File model = new File(DAO.model_watch_dir, model_name);
-        String group_name = call.get("group", "Knowledge");
+        String group_name = call.get("group", "All");
         String language_name = call.get("language", "en");
         JSONArray jsonArray = new JSONArray();
         JSONObject json = new JSONObject(true);


sudheesh001:
Can you please add a test server and test API endpoint link for the same.

Akshat-Jain:
@sudheesh001 I've been trying since morning to deploy the Server on Heroku, but I'm facing some problems. Here's the [link](https://enigmatic-brushlands-65814.herokuapp.com/) to the deployment. It seems deployed but it is showing empty responses for any query, and endpoints are also not working fine. I'm still trying to figure out what's wrong.

I'm still trying to fix the Heroku deployment, but till then, I think the following should suffice for your review. 
JSON outputs for `/cms/getSkillList.json` endpoint -
[Current JSON output](https://pastebin.com/VxFf10gT)
[JSON output as per changes done in this PR](https://pastebin.com/RGmhfmik)

I'm sorry for the inconvenience. Please take a look at the above links. Thanks :-)

Akshat-Jain:
@AnupKumarPanwar I agree that currently most of them use default group as "Knowledge". But after the implementation of group "All", I think we should shift to making it the default group.
@sudheesh001 Please give your opinions too regarding this. Thanks :-)

Akshat-Jain:
@sudheesh001 Please review. Thanks :-)

src/ai/susi/server/api/cms/ListSkillService.java
```diff
@@ -43,7 +43,7 @@ public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Aut
 
         String model_name = call.get("model", "general");
         File model = new File(DAO.model_watch_dir, model_name);
-        String group_name = call.get("group", "Knowledge");
+        String group_name = call.get("group", "All");
```
sudheesh001: Does this return all items when the `Query.group` value associated in the call is empty? If so, this could potentially become problematic as the number of skills increase. It'd always be good to send all but in a paginated fashion, maybe only display the top 30-40 items and when the client explicitly requests a reload, then load the next 30-40 etc..,

So far, I don't see a problem with this but it might need to be fixed.

Akshat-Jain: @sudheesh001 I see your point. I guess we can implement that in future PRs :-)

Akshat-Jain:
@sudheesh001 Can I have this PR merged please ? I'll implement the pagination of the JSON output for this endpoint in future PRs. Thanks :-)

src/ai/susi/server/api/cms/ListSkillService.java
sudheesh001: Sure.

Issue #774, opened 2018-05-31
cms/getSkillList.json API returns null rating on some skills


AnupKumarPanwar:
**Actual Behaviour**

cms/getSkillList.json API returns null rating on some skills and also it doesn't returns stars.

**Expected Behaviour**

The API should return 0 positive and negative rating as well as the star ratings instead of null.

**Steps to reproduce it**

1. https://api.susi.ai/cms/getSkillList.json
2. See skill_rating object of "Day in History" skill

**LogCat for the issue**
NA

**Screenshots of the issue**

![image](https://user-images.githubusercontent.com/10573038/40796013-58eb4c5e-6521-11e8-8b22-b00e0f425a18.png)

**Would you like to work on the issue?**

Yes

PR #775, opened 2018-06-01
Return star ratings in getSkillList API

AnupKumarPanwar:
Fixes
cms/getSkillList.json API returns null rating on some skills #774 

Changes: 
Modified SusiSkill.getSkillMetadata() function to return 0 count in positive and negative feedback along with the star ratings instead of null.

Screenshots for the change: 
image


src/ai/susi/mind/SusiSkill.java
```diff
@@ -525,12 +525,42 @@ public static JSONObject getSkillMetadata(String model, String group, String lan
...
+                            JSONObject newFiveStarRating=new JSONObject();
+                            newFiveStarRating.put("one_star", "0");
+                            newFiveStarRating.put("two_star", "0");
+                            newFiveStarRating.put("three_star", "0");
+                            newFiveStarRating.put("four_star", "0");
+                            newFiveStarRating.put("five_star", "0");
+                            newFiveStarRating.put("avg_star", "0");
+                            newFiveStarRating.put("total_star", "0");
+
+                            skillName.put("stars", newFiveStarRating);
```
sudheesh001: While this is nice, I am not sure if having `string` as the value of the map is a good idea. Can this instead be `int/Integer` values accross the `one_star`...... etc.., This also makes it easy to do mathematical operations on the items in these stars.

src/ai/susi/mind/SusiSkill.java
```diff
@@ -525,12 +525,42 @@ public static JSONObject getSkillMetadata(String model, String group, String lan
...
+                        JSONObject newFiveStarRating=new JSONObject();
+                        newFiveStarRating.put("one_star", "0");
+                        newFiveStarRating.put("two_star", "0");
+                        newFiveStarRating.put("three_star", "0");
+                        newFiveStarRating.put("four_star", "0");
+                        newFiveStarRating.put("five_star", "0");
+                        newFiveStarRating.put("avg_star", "0");
+                        newFiveStarRating.put("total_star", "0");
+
+                        newRating.put("stars", newFiveStarRating);
```
sudheesh001: Same feedback here. Please consider using integers as the value for each key in the object. These are mathematical and would be nice to have that type of computation being done here.

AnupKumarPanwar: I've followed the convention that is used in the thumbs feedback system. It stores the negative and position feedback count as string. So I think we should go ahead with this convention only. Also parseInt() is used at every point where the value has been accessed.

Issue #777, opened 2018-06-02
Incorrect way to use access_token


AnupKumarPanwar:
As pointed out by @dynamitechetan on susi_server gitter channel, "we are getting access_token in the request. Its wrong to do so. If the user is authenticated and the base user role is set to USER, we need not ask access_token from the client side. We can use authorization.getIdentity() method to get the user identity. "

**Would you like to work on the issue?**
Yes

PR #778, opened 2018-06-02
Fixed Incorrect way to use access_token #777

AnupKumarPanwar:
Fixes 
Fixed Incorrect way to use access_token #777 

Changes: 
Removed unnecessary getting of access_token from the request. 

Screenshots for the change: 
NA

### DIFF
--- a/src/ai/susi/server/api/cms/FiveStarRateSkillService.java
+++ b/src/ai/susi/server/api/cms/FiveStarRateSkillService.java
@@ -36,7 +36,7 @@ import java.sql.Timestamp;
  * This Endpoint accepts 5 parameters. model,group,language,skill,rating.
  * rating can be positive or negative
  * before rating a skill the skill must exist in the directory.
- * http://localhost:4000/cms/fiveStarRateSkill.json?model=general&group=Knowledge&skill=aboutsusi&stars=3
+ * http://localhost:4000/cms/fiveStarRateSkill.json?model=general&group=Knowledge&skill=aboutsusi&stars=3&access_token=6O7cqoMbzlClxPwg1is31Tz5pjVwo3
  */
 public class FiveStarRateSkillService extends AbstractAPIHandler implements APIHandler {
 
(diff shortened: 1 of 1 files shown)


AnupKumarPanwar:
@Akshat-Jain Kindly try to understand the code. Earlier the access_token was being fetched from the get parameters and then it was being used to only check if it's null then throw exception. When it's already mentioned that the base role is user then we need not make such checks. Moreover the code is no where restricting the CMS to send access_token in the URL.

Akshat-Jain:
@AnupKumarPanwar Thanks for updating the sample URL. That was exactly what I was talking about. Thanks :-)

akshatnitd:
@mariobehling @Orbiter Can we have it merged?

src/ai/susi/server/api/cms/FiveStarRateSkillService.java
```diff
@@ -70,17 +70,12 @@ public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Aut
         String skill_name = call.get("skill", null);
         File skill = SusiSkill.getSkillFileInLanguage(language, skill_name, false);
         String skill_stars = call.get("stars", null);
-        String access_token = call.get("access_token", null);
```
sudheesh001: The relevant string containing the sample query `http://localhost:4000/cms/fiveStarRateSkill.json?model=general&group=Knowledge&skill=aboutsusi&stars=3&access_token=6O7cqoMbzlClxPwg1is31Tz5pjVwo3` needs to be updated in this file now that the access_token is removed.

AnupKumarPanwar: Sample endpoint remains the same, just the nullity check on access_token in get parameters has been removed.

AnupKumarPanwar: @sudheesh001 Do you agree with my comment?

sudheesh001: Agree with this.

Issue #792, opened 2018-06-04
API for fetching last 7 days skill usage data


AnupKumarPanwar:
Making an API to fetch last 7 days skill usage data from from data/skill_rating/fiveStarSkillRating.json

**Would you like to work on the issue?**

Yes

PR #793, opened 2018-06-04
API to fetch skill usage Data

AnupKumarPanwar:
Fixes 
API for fetching last 7 days skill usage data #792

Changes: 
Add GetSkillUsageService.java to return last 7 days usage data.

Endpoint : /cms/getSkillUsage.json

Parameters
- model
- group
- language
- skill

```
Sample endpoint

/cms/getSkillUsage.json?model=general&group=Knowledge&language=en&skill=aboutsusi&callback=pc&_=1525446551181

```
Screenshots for the change: 
image

### DIFF
--- a/src/ai/susi/DAO.java
+++ b/src/ai/susi/DAO.java
@@ -112,6 +112,7 @@ public class DAO {
     //CMS Schema for server usage
     public static JsonTray skillRating;
     public static JsonTray fiveStarSkillRating;
+    public static JsonTray skillUsage;
     public static JsonTray feedbackSkill;
 
     static {
@@ -173,7 +174,7 @@ public class DAO {
             susi.addWatchpath(system_skills_localmode);
             susi.addWatchpath(susi_generic_skills_media_discovery);
         }
-                
+
         // initialize the memory as a background task to prevent that this blocks too much
         new Thread() {
             public void run() {
@@ -261,6 +262,15 @@ public class DAO {
         fiveStarSkillRating = new JsonTray(fiveStarSkillRating_per.toFile(), fiveStarSkillRating_vol.toFile(), 1000000);
         OS.protectPath(fiveStarSkillRating_per);
         OS.protectPath(fiveStarSkillRating_vol);
+
+        // Skill usage storage
+        Path susi_skill_usage_dir = dataPath.resolve("skill_usage");
+        susi_skill_usage_dir.toFile().mkdirs();
+        Path skillUsage_per = susi_skill_usage_dir.resolve("skillUsage.json");
+        Path skillUsage_vol = susi_skill_usage_dir.resolve("skillUsage_session.json");
+        skillUsage = new JsonTray(skillUsage_per.toFile(), skillUsage_vol.toFile(), 1000000);
+        OS.protectPath(skillUsage_per);
+        OS.protectPath(skillUsage_vol);
 
         //Feedback Skill storage
         Path feedbackSkill_per = susi_skill_rating_dir.resolve("feedbackSkill.json");
--- a/src/ai/susi/server/api/cms/GetSkillUsageService.java
+++ b/src/ai/susi/server/api/cms/GetSkillUsageService.java
@@ -1,0 +1,107 @@
+/**
+ *  GetSkillUsageService
+ *  Copyright by Anup Kumar Panwar, @anupkumarpanwar
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2.1 of the License, or (at your option) any later version.
+ *
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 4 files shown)


src/ai/susi/mind/SusiArgument.java
```diff
@@ -236,14 +239,94 @@ public SusiThought finding(String client, SusiLanguage language, SusiMind... min
...
+                        modelName.put(group_name, groupName);
+                        sevenDaysSkillUsage.put(model_name, modelName, true);
+                        return;
+
+                    }
+                }
+            }
+        }
+        JSONArray usageData=new JSONArray();
+        JSONObject todayUsage=new JSONObject();
```
sudheesh001: Nit: Inconsistency in indentation around `=`

src/ai/susi/DAO.java
```diff
@@ -110,6 +110,7 @@
     public static JsonTray group;
     public static JsonTray skillRating;
     public static JsonTray fiveStarSkillRating;
+    public static JsonTray sevenDaysSkillUsage;
```
sudheesh001: I think instead of calling it `sevenDays...` explicitly, maybe it's a good idea to just call it `SkillUsage` where the default parameter retrieves 7 days. Otherwise, the user should be allowed to query the duration for the skill usage. That'd make this generic and nicer.

AnupKumarPanwar: Hmm but I thought  of keeping the stats of just last 7 days, otherwise a lot of data will be generated. We can extend it later, but my suggestion to first test with limited data on the production server, so that if any bug is encountered then its easy to fix it. Still I would like to have more views on this.

Orbiter: Please make it generic so we can scale as it fits.
If we do any changes in the time frame then the name of the API is inappropriate. We would never go through all client applications and change this and we would be stuck in inappropriate naming.

Orbiter:
My approval is under the condition that @sudheesh001 argument is valid. Please change the naming.

src/ai/susi/DAO.java
AnupKumarPanwar: Ok, got your point. Sending a commit.

src/ai/susi/mind/SusiArgument.java
```diff
@@ -226,24 +229,97 @@ public SusiArgument addSkill(final SusiSkill.ID skillid) {
...
+                        skillUsage.put(model_name, modelName, true);
+                        return;
+
+                    }
+                }
+            }
+        }
+        JSONArray usageData = new JSONArray();
+        JSONObject todayUsage = new JSONObject();
+        String today=LocalDate.now().toString();
```
sudheesh001: Nit: Indentation/spacing issues around the `=`

src/ai/susi/server/api/cms/GetSkillUsageService.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        result.put("skill_name", skill_name);
+        result.put("accepted", false);
+        result.put("message", "Skill has not been used yet");
+
+        return new ServiceResponse(result);
+    }
+}
```
sudheesh001: Nit: New line at the end of file.

src/ai/susi/server/api/cms/GetSkillUsageService.java
```diff
@@ -0,0 +1,103 @@
...
+        File language = new File(group, language_name);
+        String skill_name = call.get("skill", null);
+        File skill = SusiSkill.getSkillFileInLanguage(language, skill_name, false);
+
+        JSONObject result = new JSONObject();
+        result.put("accepted", false);
+        if (!skill.exists()) {
+            result.put("message", "Skill does not exist");
+            return new ServiceResponse(result);
+
```
sudheesh001: Nit: Extra line break.

src/ai/susi/mind/SusiArgument.java
```diff
@@ -226,24 +229,97 @@ public SusiArgument addSkill(final SusiSkill.ID skillid) {
...
+
+                        if (!dateExists) {
+                            JSONObject todayUsage = new JSONObject();
+                            todayUsage.put("date", today);
+                            todayUsage.put("count", "1");
+                            usageData.put(todayUsage);
+                        }
+
+
+
```
sudheesh001: Nit: Extra line breaks.

src/ai/susi/DAO.java
```diff
@@ -259,6 +260,15 @@ public void run() {
         OS.protectPath(fiveStarSkillRating_per);
         OS.protectPath(fiveStarSkillRating_vol);
 
+        //Last 7 days skill usage storage
+        Path susi_skill_usage_dir = dataPath.resolve("skill_usage");
+        susi_skill_usage_dir.toFile().mkdirs();
+        Path skillUsage_per = susi_skill_usage_dir.resolve("skillUsage.json");
+        Path skillUsage_vol = susi_skill_usage_dir.resolve("skillUsage_session.json");
```
sudheesh001: Looks good, but how do we know it's only `7` days? Is that available in the `skillUsage`? Would it be possible to add a query parameter in the usage service for the `range`? What if I want the last 10 days? How would this have to change?

PR #804, opened 2018-06-07
Fixes #660: No reply to "?" message

### LINKED ISSUE #660: No reply to "?" message
Actual Behaviour

Susi doesn't reply to the message "?"

Expected Behaviour

It should reply by answering with a message like "Sorry I didn't get you." or repeat the previous answer in a better way

Steps to reproduce it
Changing the server file answer by adding a condition to answer "?" message according to previous answer.

LogCat for the issue

NA

Screenshots of the issue

susi3

Would you like to work on the issue?

Yes I would like to work on the issue with due  guidance.
@Orbiter

SanjanaSanikommu:
Fixes #660
Changes: 
This issue is referenced to SUSI.AI Chat.
Someone fixed #660 and necessary changes are made when user enters "?"   

Screenshots for the change: 
Before editing susi_skill
old
After editing susi_skill
question


sudheesh001:
This patch contains a lot of older commits and has been opened against master. Could you please, reopen this patch requesting merge to development instead of master branch.

PR #812, opened 2018-06-08
Add API to upload images for botbuilder

### LINKED ISSUE #806: Create API for storing custom theme images for botbuilder
**Actual Behaviour**

No API present for storing images (background image, bot icon image) required for custom design of the SUSI chatbot plugin.

**Expected Behaviour**

Develop an API for storing image.

Yes.

hkedia321:
Fixes #806 

Changes: 
- Add a new route `/cms/uploadImage.json`. This route accepts a post request consisting of three parameters:
  - `access_token`
  - `image_name`
  - `image` (the image file)

It saves the image in `susi_skill_data/botbuilder/<userEmail>/images/<image_name>`.
For example: `susi_skill_data/botbuilder/harshitkedia32@gmail.com/images/logo.png`.
The email address of the user is used to differentiate between images of the same name uploaded by different users. If we store all the images in a single folder and if two users upload different images with the same name (eg. logo.png), then one file would be overwritten by the another. Hence to store them in different places, I am using the user's email id to create different folders.

### DIFF
--- a/src/ai/susi/DAO.java
+++ b/src/ai/susi/DAO.java
@@ -91,7 +91,7 @@ import org.apache.log4j.PatternLayout;
 public class DAO {
 
     private final static String ACCESS_DUMP_FILE_PREFIX = "access_";
-    public  static File conf_dir, bin_dir, html_dir, data_dir, susi_chatlog_dir, susi_skilllog_dir, model_watch_dir, susi_skill_repo, deleted_skill_dir;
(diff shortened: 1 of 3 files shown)


AnupKumarPanwar:
Agree with @akshatnitd

hkedia321:
Yeah, I agree storing the emails publicly is not a good idea. I will be back at this with a new approach or modify this approach after discussing with mentors.

PR #816, opened 2018-06-09
Add private skill option while storing skills

### LINKED ISSUE #815: Add a parameter of "skill type" in skill data for each skill
**Actual Behaviour**

Currently, all the skills are public.

**Expected Behaviour**

In order to allow user to create skills for his/her web bot, user has to create private skills. These skills will be applicable only to the chatbot of the user.

There would be two types of skills-
- `Public` skills
- `Private` skills

Yes.

DNS-404:
Fixes #815

Changes: Added a parameter `private`. It's a boolean which is `true` (yes) if the skill is private and `false` (no) if the skill is public.


src/ai/susi/mind/SusiSkill.java
```diff
@@ -511,6 +518,7 @@ public static JSONObject getSkillMetadata(String model, String group, String lan
                     skillid.hasName(skillname)) {
 
                 skillMetadata.put("skill_name", skill.getSkillName() ==null ? JSONObject.NULL: skill.getSkillName());
+                skillMetadata.put("private_skill", skill.getPrivateSkill());
```
hkedia321: Can you try adding null check here as done for other parameters?

DNS-404: It's a `boolean` with initial value `false`. Null checks are for strings.

DNS-404:
@hkedia321 @Akshat-Jain @PrP-11 @anshumanv @akshatnitd Please review.

src/ai/susi/mind/SusiSkill.java
```diff
@@ -299,6 +301,10 @@ else if (bang_type.equals("console")) {
                     if(skillName.length() > 0)
                         json.put("skill_name",skillName);
                 }
+                if (line.startsWith("::private_skill") && (thenpos = line.indexOf(' ')) > 0) {
```
Orbiter: just call it "private", not "private_skill" because all modifiers change status of the skill somehow.

DNS-404: Okay

DNS-404:
@Orbiter Done the changes. Please review

DNS-404:
Created PR https://github.com/fossasia/susi_server/pull/855 after doing changes.

Issue #592, opened 2017-10-05
API to edit user's email


akshatnitd:
**Actual Behaviour**

Currently, there isn't any way to change the email of a user.

**Expected Behaviour**

An API exposed, where user can change the email id. It needs to done, keeping in mind the requirements of fossasia/chat.susi.ai#839

The issue can be assigned to someone else.

DravitLochan:
@akshatnitd Please edit the issue accordingly

DravitLochan:
ok. I'll work on this. @Orbiter @mariobehling Does this really needs to be implemented?

akshatnitd:
SInce, no acivity. I would like to work on it.

@Orbiter @mariobehling I would like to propose an approach. We can assign an id as a primary key to each user. It can be an integer. And then, email can be changed thereafter in the accounting model. But, many other endpoints need to be changed accordingly. Also, the existing data file should be modified accordingly. Should I go ahead with it?

akshatnitd:
For implementing email change feature, we need to change the authentication.json, authorization.json, accounting.json files on the server and change the id for existing users. How shall we go about doing it?

Orbiter:
I admit that identifying the user with the email address was a bad decision. Yes a random-like hash would be better.
As I remember, we have identifiers which are prefixed with "email:" to identify that the user id is an email. We could still follow that schema and replace the email identifier with another id like "id:<hash>". Then the original email address must be stored as well. Storage would take place only in one file: authentication ("who is the user"). To make it possible that email addresses within id-Objects are searchable, they would have to be indexed. That requires that the JsonTray Class would have to provide that. We could introduce a "@index" object which contains all searchable items there. We would also have to add a "user" object which must contain an array of user identifications which again would contain email addresses and such. We should also consider to also store old email addresses together with the attribute "deleted". Or should we actually delete old email addresses?

mariobehling:
> Or should we actually delete old email addresses?

We should store those emails in case an account gets hacked this info is important to be able to reset an account in future.

mariobehling:
I suggest an ID number. Please implement this as a unique identifier.

mariobehling:
Related to https://github.com/fossasia/susi_server/issues/837

mariobehling:
@akshatnitd As the UID is implemented please follow up.

akshatnitd:
@Orbiter Can you please guide me on how shall I go about implementing it ?

PR #823, opened 2018-06-12
Create API to store images locally

### LINKED ISSUE #806: Create API for storing custom theme images for botbuilder

hkedia321:
Fixes #806 

Changes: 
- create API `/cms/uploadImage.json` to upload image. This API accepts three parameters in a POST request:
  - access_token
  - image_name
  - image (the image file)

It stores the image in local storage inside the `data/image_uploads/<user_id>` directory. 
Later, another API will be created to retrieve this stored image (#879)

Output of successful request:
```
{
    "image_path": "963e84467b92c0916b27d157b1d45328/1529692996805_susi icon.png",
    "accepted": true
}
```
Storage location of the image:
image

Where, the number `963e84467b92c0916b27d157b1d45328` is the user id of the user and `1529692996805` is generated by getting the System time in milliseconds. This helps to differentiate between images of same name uploaded by the same user. 
The client will get the `image_path` which can be stored in another location (say `userSettings`).
Then another API will be created to retrieve the image when the client sends this `image_path` (#879).

### DIFF
--- a/src/ai/susi/SusiServer.java
+++ b/src/ai/susi/SusiServer.java
@@ -509,6 +509,7 @@ public class SusiServer {
                 GetSkillDataUrl.class,
                 UndoDeleteSkillService.class,
                 CheckRegistrationService.class,
+                UploadImageService.class,
 
                 // monitoring services
                 MonitorQueryService.class,
--- a/src/ai/susi/server/api/cms/UploadImageService.java
+++ b/src/ai/susi/server/api/cms/UploadImageService.java
@@ -1,0 +1,172 @@
+package ai.susi.server.api.cms;
+
+import ai.susi.DAO;
+import ai.susi.json.JsonObjectWithDefault;
+import ai.susi.mind.SusiSkill;
+import ai.susi.server.APIHandler;
+import ai.susi.server.AbstractAPIHandler;
+import ai.susi.server.Authentication;
+import ai.susi.server.Authorization;
+import ai.susi.server.ClientCredential;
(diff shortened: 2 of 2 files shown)


hkedia321:
@Orbiter I have now made new folders for each user, so images will be stored in  `image_uploads/<user_email>/1528807028466_logo.png`. The random integer is generated by getting the system time so as to avoid overwriting of files with common names.

hkedia321:
@anshumanv @akshatnitd @Akshat-Jain @AnupKumarPanwar @DNS-404 Please review.

hkedia321:
@mariobehling @Orbiter currently there is no id for each user, as discussed in #592. So the only way to uniquely identify the user currently seems to be their email id.

prateekiiest:
How about the specific ID be generated based on their birthdate?
@mariobehling @hkedia321

Orbiter:
It's bad but we should clarify https://github.com/fossasia/susi_server/issues/592 first, agree on a new identification method and then use that ID as part of the path.
On hold.

hkedia321:
@Orbiter @mariobehling I have now added `userId` instead of `userEmail`. Please review. Thanks.

hkedia321:
@AnupKumarPanwar updated the description with the storage location screenshot. Please check.

Orbiter:
@mariobehling The email path has been replaced by the new uuid.

PR #839, opened 2018-06-15
Add API to check if an email has signed up for SUSI.AI (Fixed #826)

### LINKED ISSUE #826: Make an API for check if user is already registered
**Actual Behaviour**
Not present

**Expected Behaviour**
An API for check if user is already registered.

Ref: fossasia/susi_android#1392
Yes

arundhati24:
Fixes #826 

Changes: Added an API to check if an email has already been registered for SUSI.AI.

```
{
  "check_email": "abc@email.com",
  "session": {"identity": {
    "type": "host",
    "name": "127.0.0.1_356778ca",
    "anonymous": true
  }},
  "exists": true,
  "accepted": true
}
```


src/ai/susi/server/api/aaa/CheckRegistrationService.java
```diff
@@ -0,0 +1,89 @@
+/**
+ *  SignUpServlet
```
akshatnitd: Can you please rename the Servlet name here.

src/ai/susi/server/api/aaa/CheckRegistrationService.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * This endpoint accepts 1 parameter email.
```
akshatnitd: Change email to check_email

src/ai/susi/server/api/aaa/CheckRegistrationService.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+		// check if id exists already
+
+		ClientCredential credential = new ClientCredential(ClientCredential.Type.passwd_login, checkEmail);
+		Authentication authentication = DAO.getAuthentication(credential);
+
+		if (authentication.getIdentity() != null) {
+			result.put("exists", true);
+		}else {
```
akshatnitd: Add space after `}`

src/ai/susi/server/api/aaa/CheckRegistrationService.java
Orbiter: The account identification is not necessarily email only. The account ID has generally an email prefix but it can be differently. Keep a generic approach

arundhati24: @Orbiter As we use email address and password as the login credentials, there was a need to check if an email address has signed up for SUSI or not. This is needed to improve the auth flow and to make it similar to Eventbrite. The aim is to ask the user to enter an email address and then check whether that email address has already been signed up for SUSI or not. If that email address has already signed up, we would simply show the password field otherwise we would display the sign up screen with the email address already filled in. As per my understanding, we do not have any identification for an account unless it has signed up for SUSI, therefore, I kept only one parameter i.e. 'email address'.
It would be really very helpful if you could please reflect a bit upon what other parameters should be taken as per the requirement described above and also if there is something that I have missed. Thanks.

arundhati24:
@Orbiter It would be helpful if you could please check my reply to your review and suggest necessary changes accordingly as this is a blocker to the improvement in auth flow in the Android client. Thanks. :)

PR #841, opened 2018-06-16
Generate unique id for user while sign up

### LINKED ISSUE #837: Generate user id for each user
**Actual Behaviour**

No user id present for each user. Only user email is used to identify the user.

**Expected Behaviour**

Unique user id should be generated and assigned to each user when they sign up. This user id would be used in places where email id can't used for privacy reasons (like storing images, for bot token, etc). Also email can change, but user id would be permanent.

Yes, along with @akshatnitd and @AnupKumarPanwar.

hkedia321:
Fixes #837 

Changes: 
- In `ClientIdentity.java` put a new identity type `id`. 
- In `Authentication.java` make a function `setEmail` to save email of logged in users.
- In `SignUpService.java` generate unique id by using `UUID.randomUUID()`. Store this in the `id` field of the user. Also store the email of the user in `email` field.
- In `LoginService.java` send the generated `id` in the response.
- Since `identity.getName()` will now return the id, use `authentication.getString("email")` where user email is expected.

*Script to run which assigns user id to past users:* [link](https://github.com/hkedia321/susi_server/blob/scriptGenerateUniqueIdForPastUser/src/ai/susi/server/api/aaa/AssignUserIdPastUsers.java)

Screenshots for the change: 
The user object (for new user) in `authentication.json` is now:
```
"passwd_login:testing@gmail.com": {
    "salt": "KwhPQWhIHtjuAclzDQbt",
    "id": "id:364dc634181a43c6b6ef7df3111b123d",
    "email": "testing@gmail.com",
    "passwordHash": "sS2MEAoXMC4G3tD4HReFoHPLT6Q3FOyEHeJ8saJ6P6w=",
    "activated": true
  }
```
Successful Login response:
```
{
  "accepted": true,
  "valid_seconds": 604800,
  "access_token": "esryw2s6gWBd8Gl0HrMVZpHRwTrUSY",
  "id": "364dc634181a43c6b6ef7df3111b123d",
  "message": "You are logged in as testing@gmail.com",
  "session": {"identity": {
    "type": "host",
    "name": "0:0:0:0:0:0:0:1_af2c1fe3",
    "anonymous": true
  }}
}
```
In `authentication_session.json`:
```
"access_token:b2aQ1vSrqzUxRFm9ainPmbMetKsp4k": {
    "expires_on": 1529930028,
    "id": "id:364dc634181a43c6b6ef7df3111b123d",
    "email": "testing@gmail.com"
  }
```


hkedia321:
@akshatnitd can you please specify what kind of changes we need to make in the APIs. As I see, the `ClientIdentity` will now use `id` instead of `email`. So I have replaced `identity.getName()` with `authentication.getString("email")` where userEmail is expected. In other places, `id` seems fine.
Also, we need to run a script to assign user id to all the past users. For that, I have made a [script here](https://github.com/hkedia321/susi_server/blob/a61bf7a5239c7c44b32ddb98a3418a402a31a0b3/src/ai/susi/server/api/aaa/AssignUserIdPastUsers.java). This can be run on the server to update past user's ids. Thanks.

hkedia321:
@akshatnitd I have made the changes in the related APIs. I have replaced `authorization.getIdentity().getName()` with `authentication.getString("email")` logic. Now they will correctly return `email` as expected instead of `id`. Can you please review again.

hkedia321:
Already Solved in c212cb70d17ab2e881558fd034412683f2de27fb

Issue #842, opened 2018-06-17
Skill sorting by "most used first" order


AnupKumarPanwar:
Store the usage count in skill metadata and sort the skills in decreasing order of their usage.

**Would you like to work on the issue?**

Yes

PR #849, opened 2018-06-18
Sorted skills by Usage

AnupKumarPanwar:
Fixes 
Skill sorting by "most used first" order #842 

Changes: 

Sorting parameters :
- filter_type=usage
- filter_name=ascending | descending
- duration (by default 7) (updated)

```
Sample endpoint (updated)

/cms/getSkillList.json?model=general&group=All&language=en&applyFilter=true&filter_name=descending&filter_type=usage&duration=30
```

Screenshots for the change: (updated)
image


src/ai/susi/mind/SusiSkill.java
```diff
@@ -547,65 +567,60 @@ public static JSONObject getSkillMetadata(String model, String group, String lan
...
+            JSONObject modelName = skillUsage.getJSONObject(model);
+            if (modelName.has(group)) {
+                JSONObject groupName = modelName.getJSONObject(group);
+                if (groupName.has(language)) {
+                    JSONObject languageName = groupName.getJSONObject(language);
+                    if (languageName.has(skillname)) {
+                        JSONArray skillName = languageName.getJSONArray(skillname);
+
+                        // Fetch only last seven days skill usage data for sorting purpose.
+                        int duration = 7;
```
Orbiter: Why is this not generic? Please pass the '7' as an argument of the getSkillUsage call

src/ai/susi/server/api/cms/GetSkillUsageService.java
```diff
@@ -86,8 +86,8 @@ public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Aut
                         JSONArray skillUsage = languageName.getJSONArray(skill_name);
                         result.put("skill_name", skill_name);
```
Orbiter: ass the duration time for the skill usage sort timeframe here

src/ai/susi/mind/SusiSkill.java
```diff
@@ -547,65 +567,60 @@ public static JSONObject getSkillMetadata(String model, String group, String lan
...
-                skillMetadata.put("skill_rating", newRating);
-            }
         }
+        JSONObject newRating=new JSONObject();
+        newRating.put("negative", "0");
+        newRating.put("positive", "0");
+        newRating.put("feedback_count", 0);
+
+        JSONObject newFiveStarRating=new JSONObject();
+        newFiveStarRating.put("one_star", 0);
```
Orbiter: very strange data structure. Who created this data structure by naming of the stars? You could have used an JSONArray containing integers for the five stars.

src/ai/susi/mind/SusiSkill.java
```diff
@@ -500,6 +500,7 @@ public static JSONObject getSkillMetadata(String model, String group, String lan
         skillMetadata.put("dynamic_content", false);
         skillMetadata.put("examples", JSONObject.NULL);
         skillMetadata.put("skill_rating", JSONObject.NULL);
+        skillMetadata.put("week_usage", 0);
```
Orbiter: Is it necessary to add this to the skill metadata? This is not good in a concurrent environment. Different calls will modify the metadata all the time. Furthermore: naming this "week_usage" is not a generic way. If the usage is sorted in a generic way a field should not be named as such.

src/ai/susi/mind/SusiSkill.java
```diff
@@ -520,10 +521,29 @@ public static JSONObject getSkillMetadata(String model, String group, String lan
                 skillMetadata.put("terms_of_use", skill.getTermsOfUse() ==null ? JSONObject.NULL:skill.getTermsOfUse());
                 skillMetadata.put("dynamic_content", skill.getDynamicContent());
                 skillMetadata.put("examples", skill.getExamples() ==null ? JSONObject.NULL: skill.getExamples());
+                skillMetadata.put("skill_rating", getSkillRating(model, group, language, skillname));
+                skillMetadata.put("week_usage", getSkillUsage(model, group, language, skillname));
```
Orbiter: see above

AnupKumarPanwar:
@Orbiter Please review again. Made the required changes.

AnupKumarPanwar:
@Orbiter yes I agree, we can have a better data structure for the 5 star rating storage. But I also think that there may not be much amendments in its structure as it only stores the summary of the ratings. Moreover the current structure allows us to have heterogeneous nature which would not have been possible in Array.

PR #860, opened 2018-06-19
Fixes #858: Add count param in ListSkillService.java

### LINKED ISSUE #858: Add option to accept count=x param to return Top x skills after filtering
**Actual Behaviour**
Feature not present

**Expected Behaviour**
* Add option to accept count=x param to return Top x skills after filtering

**Steps to reproduce it**
NA

Yes

akshatnitd:
Fixes #858 

* Add count param in ListSkillService.java
* Will be used in the home page of Skill CMS to suppose display Top 5 skills, etc

Screenshots for the change: 
NA

### DIFF
--- a/src/ai/susi/server/api/cms/ListSkillService.java
+++ b/src/ai/susi/server/api/cms/ListSkillService.java
@@ -16,6 +16,7 @@ import java.util.*;
 /**
  * This Servlet gives a API Endpoint to list all the Skills given its model, group and language.
  * Can be tested on http://127.0.0.1:4000/cms/getSkillList.json
+ * Other params are - applyFilter, filter_type, filter_name, count
  */
 
 public class ListSkillService extends AbstractAPIHandler implements APIHandler {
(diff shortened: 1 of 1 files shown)


src/ai/susi/server/api/cms/ListSkillService.java
```diff
@@ -48,6 +49,22 @@ public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Aut
         JSONArray jsonArray = new JSONArray();
         JSONObject json = new JSONObject(true);
         JSONObject skillObject = new JSONObject();
+        String countString = call.get("count", null);
```
AnupKumarPanwar: You get get count here directly. call.get() returns the the data type as per the default value.

AnupKumarPanwar:
The functioning looks good, but another approach can be to initialise count to the passed value or jsonArray.length based on whether to apply countFilter or not and update the loop parameters accordingly. Rest LGTM.

Issue #866, opened 2018-06-20
The Five Star Skill Rating JSON file is not getting stored properly


AnupKumarPanwar:
**Actual Behaviour**

The five start skill rating API is sometimes over writing the previous data in fiveStarSkillRating.json. Also some times multiple copies of same JSON Object is being created. This happens when multiple users rate the same skill. Though the overall rating is not affected, but the rating by each individual is overwritten.

**Expected Behaviour**

The fiveStarSkillRating.json file should be updated with correct data and multiple copies should not be created.

**Steps to reproduce it**

Give different rating by different users on the same skill.

**LogCat for the issue**

NA

**Screenshots of the issue**

See the ratings on aboutsusi skill
![image](https://user-images.githubusercontent.com/10573038/41673528-d46e5984-74da-11e8-8bd6-fb946c4329cc.png)

**Would you like to work on the issue?**

Yes

Issue #864, opened 2018-06-20
The Skill usage data is not getting stored properly


AnupKumarPanwar:
**Actual Behaviour**

The skill usage data and country wise skill usage data is sometimes over writing the previous data. Also some times multiple copies of same JSON Object is being created.

**Expected Behaviour**

The skill usage JSON file should be updated with correct data and multiple copies should not be created.

**Steps to reproduce it**

When we use a skill in different countries some usage data is overwritten.

**LogCat for the issue**

NA

**Screenshots of the issue**

![image](https://user-images.githubusercontent.com/10573038/41645522-e0065efa-748e-11e8-9a33-b3224ff33442.png)

**Would you like to work on the issue?**

Yes

PR #868, opened 2018-06-20
Fixed bugs in updating the usage and feedback JSON file

AnupKumarPanwar:
Fixes 
The Skill usage data is not getting stored properly #864
The Five Star Skill Rating JSON file is not getting stored properly #866

Changes: 
Earlier the same JSONObject was being used to assigned a new value in the loop which overwrites the previous values of the JSONObject with the current value. So to fix this a new JSONObject is created everytime in the loop.

Screenshots for the change: 
image


AnupKumarPanwar:
@anshuman73 @akshatnitd @akshatnitd Please review

anshuman73:
@AnupKumarPanwar , I'm not sure I'm the right person you meant to tag. Please check once.

PR #870, opened 2018-06-20
Store private skills in different repository

### LINKED ISSUE #863: Store private skills of the users in another git repository
**Actual Behaviour**

Currently, the "private" skills made by the users is published in the public [susi_skill_data](https://github.com/fossasia/susi_skill_data) repository

**Expected Behaviour**

All private skills must be published in another git repository which will not be public. The skill data will have a different structure than the existing public repo and the skills will be co related to the ids of the users who created them. 
This is corresponding to [#738#issuecomment-398121894](https://github.com/fossasia/susi_server/issues/738#issuecomment-398121894).
This will be the first step to creating private skills. The next step will include applying the private skill only to the conversations of that particular user and not to whole of SUSI.

**Steps to reproduce it**

Make a skill in the botbuilder. It will be public.

Yes

hkedia321:
Fixes #863 

Changes: 
- In `CreateSkillService.java` detect if the client sends an attribute `private`. If it sends, then store the skill in the `susi_private_skill_data` repository. Else store in normally in the `susi_skill_data` repository.
- In `SusiSkill.java` check for empty folder, otherwise it was throwing an error.

Screenshots for the change: 
directory structure in the repository:
```
└── users
    ├── user1id
    │   ├──  custom
    │       ├── Alarms\ and\ Clocks
    ├── user2id
    │   ├──  custom
    │       ├── Alarms\ and\ Clocks
```
Eg: On executing it locally, the private skill was successfully stored in:
image

### DIFF
--- a/src/ai/susi/DAO.java
+++ b/src/ai/susi/DAO.java
@@ -91,7 +91,7 @@ import org.apache.log4j.PatternLayout;
 public class DAO {
 
     private final static String ACCESS_DUMP_FILE_PREFIX = "access_";
-    public  static File conf_dir, bin_dir, html_dir, data_dir, susi_chatlog_dir, susi_skilllog_dir, model_watch_dir, susi_skill_repo, deleted_skill_dir;
+    public  static File conf_dir, bin_dir, html_dir, data_dir, susi_chatlog_dir, susi_skilllog_dir, model_watch_dir, susi_skill_repo, private_skill_watch_dir, susi_private_skill_repo, usi_skill_repo, deleted_skill_dir;
     public static String conflictsPlaceholder = "%CONFLICTS%";
     private static File external_data, assets, dictionaries;
     private static Settings public_settings, private_settings;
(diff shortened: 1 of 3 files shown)


src/ai/susi/server/api/cms/CreatePrivateSkillService.java
```diff
@@ -0,0 +1,209 @@
+package ai.susi.server.api.cms;
```
Orbiter: up to here it's fine but I don't see a requirement to make a new servlet. The private skill shall also be made with the CreateSkillService.java servlet and that servlet needs to read a 'private' attribute and then act acordingly

hkedia321: @Orbiter I have now removed the new API and made changes to the  `CreateSkillService.java` file to detect `private` attribute and then store them in `susi_private_skill_data` repository. Please review. Thanks,

PR #884, opened 2018-06-24
Fix bug with svg image

### LINKED ISSUE #883: SUSI.AI svg file is not displaying
**Actual Behaviour**

SUSI.AI svg file is present but not displaying on [api.susi.ai](api.susi.ai). 
screen shot 2018-06-25 at 1 53 04 am
This is maybe because server is serving it with an incorrect `content-type`. It's described in detail [here](https://css-tricks.com/snippets/htaccess/serve-svg-correct-content-type/).

**Expected Behaviour**

It should work. All the images are in png format. So I'll use the png image in place of svg image.
I was able to see the image when I tested the website on my local machine. Problem arrises when it's run on a web server.

Yes

DNS-404:
Fixes #883 

Changes: Using inline SVG for now.


Orbiter:
was there not a different way to fix this? The svg usually gives more flexibility and more quality.

DNS-404:
I can see the svg image if I simply open the html file but I'm not able to see it when I run the website on a web server. I'm not sure about the reason. I'm trying to find a solution to make svg files work.

Orbiter:
The `.htaccess` file belongs to an apache web server. Our server is self-contained, therefore we have a different way to care about mime types. We did not support the right type so far, the commit https://github.com/fossasia/susi_server/commit/f48fc12e663a6f636c1b2702d4c80c23f14cd18f fixes that. Please check if the problem is still there.

DNS-404:
@Orbiter Okay. This commit should have solved this issue but the content-type of image file is still being shown as `text/xml` so it's still not working.

DNS-404:
@Orbiter @mariobehling I found a solution for this. I've tried it and it's working. We can use a data URI to specify the content-type ourselves. For example -
```
<img alt="susi logo" class="susi-logo" src="data:image/svg+xml,<svg></svg>" />
```
The only problem is that I would have to write inline svg inside the image tag as shown above.

DNS-404:
It's working now. Please review.

Orbiter:
Inline svg is strange, we should find out why the server does not deliver the right mime type.

DNS-404:
@mariobehling @Orbiter Can this be merged for now because I'm not able to find why server can't deliver correct mime type? Currently, I'm using inline SVG for the SUSI.AI logo through this PR. We can change this later once the mime type problem is solved.

DNS-404:
@Orbiter I just noticed that right mime type is being delivered on server now. So, this PR is no longer required.

PR #887, opened 2018-06-25
Add API to get images from local storage

### LINKED ISSUE #879: Create API to retrieve images from local storage
**Actual Behaviour**

No API present to retrieve images stored in local storage of the server.

**Expected Behaviour**

An API must exists so that the client can send image path like `963e84467b92c0916b27d157b1d45328/1529692996805_susi icon.png` and the server will return the corresponding image. The client needs to send the full path including the user id because the base user role for the API would be anonymous since it should be accessible from areas such as the web bot where the access_token of the user would not be saved.

This would be done after #806 is solved. 

Yes.

hkedia321:
Fixes #879 

Changes: 
- Add a new servlet `GetImageServlet.java` to return the image from local storage given its path and name. 

Screenshots for the change: 
Testable at: `http://127.0.0.1:4000/cms/getImage.png?image=963e84467b92c0916b27d157b1d45328/1529935805009_background.png`

image

### DIFF
--- a/src/ai/susi/SusiServer.java
+++ b/src/ai/susi/SusiServer.java
@@ -588,6 +588,9 @@ public class SusiServer {
         servletHandler.addServlet(Sitemap.class, "/sitemap.xml");
         servletHandler.addServlet(ThreaddumpServlet.class, "/threaddump.txt");
 
+        // cms api
+        servletHandler.addServlet(GetImageServlet.class, "/cms/getImage.png");
+
         // aggregation api
         servletHandler.addServlet(GenericScraper.class, "/susi/genericscraper.json");
 
--- a/src/ai/susi/server/api/cms/GetImageServlet.java
+++ b/src/ai/susi/server/api/cms/GetImageServlet.java
@@ -1,0 +1,74 @@
+/**
+ *  GetImageService
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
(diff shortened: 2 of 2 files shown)


hkedia321:
@Orbiter I think the image fetching API should not have any user restrictions because this link should be accessible from anywhere. For example, in the SUSI web bot plugin, the background image of the chatbot can simply be fetched from `https://api.susi.ai/cms/getImage.png?image=963e84467b92c0916b27d157b1d45328/1529935805009_background.png` depending on the user's settings. This web bot will be deployed on many different websites. We cannot include `access_token` (to identify user id) here because that would make the `access_token` public. Therefore the base user role for this API should be Anonymous. 
The `UUID` was required to enhance the user's privacy. By looking at the above link, we cannot say who is the user who uploaded the image. Whereas if we used email id instead of UUID the user could easily be identified and their email id will be exposed. I think this is same logic by which AWS and other cloud storage works i.e after uploading the image we get a link ([like this](https://d1.awsstatic.com/product-marketing/organizations/AWS%20Organizations%20Video%20Thumbnail.cca0dc8029b60555111d1d7f8125f86393a102b7.png)) consisting of unique ids, which we can use anywhere to access the image. 
Also, I made the servlet following the approach of [MapServlet](https://github.com/fossasia/susi_server/blob/development/src/ai/susi/server/api/vis/MapServlet.java) and [AssetServlet](https://github.com/fossasia/susi_server/blob/development/src/ai/susi/server/api/service/AssetServlet.java).
Please let me know if this approach needs to be changed. Thanks.

Issue #859, opened 2018-06-19
Save the chatbots created by a user


DNS-404:
**Expected Behaviour**

In issue https://github.com/fossasia/susi_skill_cms/issues/744, users will be able to create multiple chatbots and these bots will be saved in the server. So, we need to:
- Identify each and every bot uniquely using a unique ID.
- Save the chatbot. (Chatbot is private/protected skills)
- Fetch these chatbots and show them to user.

The private/protected skill will be stored in the separate repository now after PR https://github.com/fossasia/susi_server/pull/870 but we need to keep track of the chatbots created by particular user. This is the purpose of this issue.

**Screenshots of the issue**

The bots will be shown in form of cards. It'll be similar to the following screenshot of [landbot.io](landbot.io).
<img width="1183" alt="screen shot 2018-06-19 at 4 01 53 pm" src="https://user-images.githubusercontent.com/31174685/41592693-04895444-73db-11e8-9d20-ba5aa122133a.png">

**Would you like to work on the issue?**

Yes

DNS-404:
@Orbiter After we name these skills as `protected` (PR - https://github.com/fossasia/susi_server/pull/855), I can generate the unique id in javascript itself for each bot and then save it along with other settings of private skills and themes in a new folder named 'Bots' in the Data directory.
Will this be a correct approach?

Orbiter:
What does define a 'bot' just a skill or something else?

DNS-404:
@Orbiter 'bot' is actually whole chatbot. Now this includes all the `private`/`protected` skills created by user for this specific bot along with the theme settings of the bot. All bots will be identified by a `name` as specified in issue fossasia/susi_skill_cms#744.

DNS-404:
@Orbiter After the previous week's meeting, we discussed some new things regarding design settings of chatbot. So here's the updated info, please ignore my above comment.
A 'bot' is just a private/protected skill. Nothing else. Now, this private/protected skill will include the skill of that bot obviously along with the design settings as well. Design settings include the background colour of chatbot, colour of text, colour of textbox etc. We want to store the hexadecimal codes of these colours in the "skill language" itself. For example, it can be stored somewhat like this:
```
::design #002ee4, #0077e5, #ff3535, #ffb0b0, #454a64, #f55757
```
These are hexadecimal codes of Body background, User's message box background, User's message text color, Bot's message box background, Bot's message text color and Bot Icon respectively.
[...]

PR #888, opened 2018-06-25
Add chatbot folder in data directory to save chatbots

DNS-404:
Partially fixes issue #859 

Changes: `chatbot.json` file gets generated at `susi_server/data/chatbot`.

**Note** - I'll create the API after the json structure is confirmed. Hence this is just a partial fix.


DNS-404:
@Akshat-Jain Done.

PR #904, opened 2018-06-28
Fixes #901: Keep duration as optional param in getSkillList.json

### LINKED ISSUE #901: The usage_count and feedback_count returns as 0 in getSkillList.json
**Actual Behaviour**
The usage_count and feedback_count returns as 0 in getSkillList.json. It is because, the duration parameter is defaulted as `0`.

**Expected Behaviour**
* Make duration as an optional parameter.

**Steps to reproduce it**
* Hit this endpoint - 

Yes

akshatnitd:
Fixes #901 

* Keep duration as optional param in getSkillList.json
* Initialised duration as `-1`

Screenshots for the change: 
NA

### DIFF
--- a/src/ai/susi/mind/SusiSkill.java
+++ b/src/ai/susi/mind/SusiSkill.java
@@ -625,7 +625,10 @@ public class SusiSkill {
                         JSONArray skillName = languageName.getJSONArray(skillname);
 
                         // Fetch skill usage data for sorting purpose.
-                        int startIndex = skillName.length() >= duration ? skillName.length()-duration : 0;
+                        int startIndex = 0;
+                        if(duration >= 0) {
+                            startIndex = skillName.length() >= duration ? skillName.length()-duration : 0;
+                        }
                         for (int i = startIndex; i<skillName.length(); i++)
                         {
                             JSONObject dayUsage = skillName.getJSONObject(i);
--- a/src/ai/susi/server/api/cms/ListSkillService.java
+++ b/src/ai/susi/server/api/cms/ListSkillService.java
@@ -52,7 +52,7 @@ public class ListSkillService extends AbstractAPIHandler implements APIHandler {
         File model = new File(DAO.model_watch_dir, model_name);
         String group_name = call.get("group", "All");
         String language_name = call.get("language", "en");
-        int duration = call.get("duration", 0);
+        int duration = call.get("duration", -1);
         JSONArray jsonArray = new JSONArray();
         JSONObject json = new JSONObject(true);
         JSONObject skillObject = new JSONObject();


src/ai/susi/mind/SusiSkill.java
```diff
@@ -625,7 +625,10 @@ public static int getSkillUsage(String model, String group, String language, Str
                         JSONArray skillName = languageName.getJSONArray(skillname);
 
                         // Fetch skill usage data for sorting purpose.
-                        int startIndex = skillName.length() >= duration ? skillName.length()-duration : 0;
+                        int startIndex = 0;
+                        if(duration >= 0) {
+                            startIndex = skillName.length() >= duration ? skillName.length()-duration : 0;
+                        }
```
AnupKumarPanwar: This will make the API complex. For every call in which duration parameter is not given the loop will run over the whole JSONArray unnecessarily.

akshatnitd: This is done, so that we can get an overall usage stats. Currently, we do not know the duration that has to be entered to get the overall stats. For duration based filtering, we can always add the duration parameter. Please correct me, if I am wrong.

AnupKumarPanwar: Suppose we don't want usage based sorting, but according to these changes it will still access the whole usage data. For fetching the whole usage data we can send a big integer in the duration parameter.

akshatnitd: @AnupKumarPanwar If we don't want usage based sorting, what should be the duration value in that case?
Also, if we send a big number in duration param, won't that be same as running the entire loop. Also, it wouldn't give us the result of the entire usage data, if it exceeds that big number. 

I agree to the time complexity issue, raised by you. But, for that, we need to make modifications in the way, we store usage data. Because, getting the highest used skills, considering the lifetime usage is an important indicator for us, wrt to the stats and analytics part.

AnupKumarPanwar: @akshatnitd If we do not want a usage based sorting then we will not send the duration parameter. According to the existing code it will make the duration=0 and startIndex = skillName.length(). Which means that the loop to count usage data will run 0 times.
But according to your PR, by default duration = -1 and startIndex=0, which means the loop will run over whole data.

akshatnitd: @AnupKumarPanwar I think you are talking about the case of `/getSkillList.json` API call. Currently, it shows a usage count of `0` for all skills, as duration isn't passed. Shouldn't we instead remove it from response, rather than showing `0` ?

akshatnitd: @AnupKumarPanwar I think you are talking about the case of `/getSkillList.json` API call. Currently, it shows a usage count of `0` for all skills, as duration isn't passed. Shouldn't we instead remove it from response, rather than showing `0` ?

AnupKumarPanwar: Yes we can remove it from the response when duration parameter isn't passed. When the client is not interested in the usage count, it will not pass duration parameter in the API call. So considering it as 0 will reduce the unnecessary complexity. We can remove it too when the duration is zero.

akshatnitd:
@hkedia321 @mariobehling Your views on it. This PR will make the duration parameter optional and let us return the skills based on overall metrics too.

@AnupKumarPanwar For, reducing the complexity of each API call, we need to implement a common counter, which stores the overall usage count and does that in O(1).

Orbiter:
@akshatnitd good point. What about that usage counter, is that complex?
Let's proceed in small steps. If we can work on with these changes, give your thumbs up. And then link next steps here.

akshatnitd:
@Orbiter I think, we can go ahead with this PR logic and return overall usage count every time the API is hit.

The next step would be to maintain an overall usage counter that updates everytime the skill is used. When duration is not sent, we need to just send that overall counter value, else we can loop through the datewise usage of skill and return the count accordingly.
Issue link : https://github.com/fossasia/susi_server/issues/938

PR #917, opened 2018-06-30
Added regex check for Mac Address in AddNewDevice.java file

### LINKED ISSUE #898: Use regex check for Mac Address in AddNewDevice.java file
**Actual Behaviour**

Currently, there is no validation on the format of the Mac Address passed as the query parameter.

**Expected Behaviour**

There should be a regex check to check if the format of Mac Address provided as the query parameter is valid or not.

Yes

Akshat-Jain:
Fixes #898 

Changes: Added a regex check for Mac Address in the AddNewDevice.java file. Now, while adding any new device, a proper Mac Address need to be passed as the query parameter.
Also corrected the indentation at some places in the file.

Screenshots for the change: 
If any invalid Mac Address is passed:
screen shot 2018-06-30 at 2 00 45 pm


Akshat-Jain:
@Orbiter Yes, it was discussed in one of the meetings that we should store the geolocation of the connected devices also along with the other information. A Map View has also been implemented on the Web Client which displays the devices on a map based on their geolocation.
However, I see your point. Should I modify the API and remove throwing of exceptions when location info is not provided, and instead store a string like "Location not available" as the default value?

mariobehling:
@Akshat-Jain 

> Should I modify the API and remove throwing of exceptions when location info is not provided, and instead use a string like "Location not available" as the default value?

Yes, sounds like a good idea. However, our goal is to show the device location on the map in the clients still.

Akshat-Jain:
@mariobehling Yes, that is exactly why I made the API to throw exception when location info is not provided. If I modify the API to use default string as something like "Location not available", how should I proceed with the Map View implementation in the Devices tab in Settings? Or should we hold on to the current implementation of the API for now?

Akshat-Jain:
@Orbiter @mariobehling Made the changes as discussed in today's meeting. Now, the API works even if the latitude or longitude info is not provided. Now, `Latitude not available` and `Longitude not available` are stored as the default strings. Will make the required modifications on the Web Client after this PR gets merged. Please review. Thanks :-)

PR #937, opened 2018-07-06
Added a new boolean parameter to skill metadata to mark review status of the skill

### LINKED ISSUE #934: Add a parameter to indicate status of skill in skill metadata
**Actual Behaviour**

Currently, there is no such parameter in the skill metadata.

**Expected Behaviour**

There should be a boolean parameter which should indicate the status of a Skill - reviewed or non-reviewed.

Yes.

Akshat-Jain:
Fixes #934 

Changes: Added a parameter `reviewed`. It's a boolean which is `true` if the skill has been reviewed and `false` otherwise.


Akshat-Jain:
@akshatnitd Thanks for pointing it out. This doesn't seem to be the correct approach to this. I think we should store this info in a separate JsonTray. Closing this PR as it is not relevant anymore. Thanks :-)

PR #951, opened 2018-07-09
Optimized docker images 

robbyoconnor:
Changes: I did some optimizations to the image. The docker images left the cache intact and used an unnecessary number of layers. This cleans up things and hopefully makes the image size better as well :) 

I changed the base image to the openjdk8 images


robbyoconnor:
This broke things -- I made an error - I'm making a PR to fix it.

PR #997, opened 2018-07-17
Added next page value in the pagination API

### LINKED ISSUE #996: Pagination API in skill listing can be better.
The pagination API in the skill listing can be updated to send the endpoint of the next page so that it is easy to implement on the client side and end condition, when to stop calling pagination API, is known. 

Yes

AnupKumarPanwar:
Fixes #996 

Changes: 
Return nextPage value with the skill list. The nextPage implies the next offset value that needs to be sent for fetching the next page of skill. Its value is null if no next page is available.

Screenshots for the change: 
NA


AnupKumarPanwar:
@Akshat-Jain but that's not so dynamic. The /aaa/getUsers.json returns 50 objects in 1 page. And as far as I understand that API first sends the total number of pages, say x and then the API is called x times. 
But in my approach, I'm sending the offset of the next page till the pages are available and send null in next page if no further pages exist. Also, the number of objects per page is dynamic. I think the implementation is easier in my approach. I would like to hear from @akshatnitd and @anshumanv .`

AnupKumarPanwar:
@Akshat-Jain @akshatnitd please review again.

src/ai/susi/server/api/cms/ListSkillService.java
```diff
@@ -454,6 +459,23 @@ public int compare(JSONObject a, JSONObject b) {
...
+            if (call.get("getPageCount", false) == true) {
+                int pageCount = jsonArray.length() % count == 0 ? (jsonArray.length() / count) : (jsonArray.length() / count) + 1;
+                json.put("pageCount", pageCount);
+                json.put("accepted", true);
+                json.put("message", "Success: Fetched count of pages");
+                return new ServiceResponse(json);
+            }
+
+            if (call.get("getSkillCount", false) == true) {
+                json.put("userCount", jsonArray.length());
```
PrP-11: @AnupKumarPanwar Please change userCount to skillCount

PrP-11: @AnupKumarPanwar Please change userCount to skillCount.

AnupKumarPanwar:
@akshatnitd Pagination should be implemented on all the clients.

AnupKumarPanwar:
@akshatnitd @akshatnitd @anshumanv How are you planning to implement the pagination? By creating numbered pages or by infinite scrolling?

Orbiter:
There are two general approaches to implement pagination:
- the client requests a page number (as implemented here) and there is a setting somewhere how large a page is
- the client request simply results with a beginning index and a count number.

The second approach gives much more flexibility while it mights be easier in some situations to store pages as with the first option.

These approaches can also be seen in the opensearch standard as described in https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#opensearch-11-parameters

From my point of view it is better to use the second aproach with a start index and a count. However if you choose here to follow the first concept so please go with it.

@AnupKumarPanwar I would welcome if you reflect these options to finalize the feature. I would like to see that both options are _options_

AnupKumarPanwar:
@Orbiter So what this PR does is, it take 2 parameters, the page number and the count (number of skills per page). The total number of pages is calculated as (total number of skills)/(count). If no count parameter is given, all the skills are returned. The beginning index is calculated from the page number itself. Is it good to go or some changes are still required in the approach.

AnupKumarPanwar:
Done @Orbiter .

AnupKumarPanwar:
@akshatnitd @anshumanv @Orbiter @mariobehling please review again.

PR #1037, opened 2018-07-21
Added top games in metrics

### LINKED ISSUE #1023: Add top games to the metrics endpoint
**Actual Behaviour**
Need to add metrics for top games.

Yes

AnupKumarPanwar:
Fixes #1023 

Changes:
Add top games to the metrics endpoint.

Screenshots for the change: 
image


AnupKumarPanwar:
@Akshat-Jain Yes, the approach is same as you described. But can you please tell how the other approach is better than the current one, like in terms of complexity or what?

Akshat-Jain:
@AnupKumarPanwar Yes, I believe it should be better. We're already looping through all the Skills in the `for` loop on line 100. We can check in this loop only if any Skill belongs to `Games, Trivia and Accessories` group and put it in another List. Then the Comparator function can be used for this List.

src/ai/susi/server/api/cms/SkillMetricsDataService.java
```diff
@@ -188,6 +188,34 @@ public int compare(JSONObject a, JSONObject b) {
...
 
+        // Get games skills based on ratings
+        Collections.sort(jsonValues, new Comparator<JSONObject>() {
+            @Override
+            public int compare(JSONObject a, JSONObject b) {
+                float valA;
+                float valB;
+                int result=0;
+
+                if (!a.get("group").toString().equals("Games, Trivia and Accessories")){
```
Orbiter: I am a bit unhappy about the fixed group name here. This method does also not work with different kind of group categories like we would have with personas or private skills. I propose to use a variable which the client can set. Please add a post variable `topGroups` where a client can pass the string "Games, Trivia and Accessories".

That would make this feature more flexible.

src/ai/susi/server/api/cms/SkillMetricsDataService.java
```diff
@@ -188,6 +188,34 @@ public int compare(JSONObject a, JSONObject b) {
...
+                    result = Float.compare(valB, valA);
+                } catch (JSONException e) {
+                    e.printStackTrace();
+                }
+                return result;
+            }
+        });
+
+        JSONArray topGamesData = getSlicedArray(jsonValues, count);
+        skillMetrics.put("topGames", topGamesData);
```
Orbiter: In the context of having more flexibility with the preferred group, the name of the metrics "topGames" does not make sense. Please rename "topGames" to "top".

src/ai/susi/server/api/cms/SkillMetricsDataService.java
Orbiter: .. and use that variable instead of the fixed group name.

Orbiter:
I modified the issue description to remove a possible confusion where this should go.
Please make the group name generic by adding a call option.
A fixed group name in the code is just strange and inacceptable from a design standpoint.

AnupKumarPanwar:
Ok I'll generalise the code.

AnupKumarPanwar:
@Orbiter I've updated my PR and its description, please review.

PR #1064, opened 2018-07-25
Apply private skill in susi chat

### LINKED ISSUE #1043: Private skill should work in a priority manner similar to dream skill
**Actual Behaviour**

Users are unable to use the private skills in the chatbots they deploy.

**Expected Behaviour**

The private skill should be applied in a priority manner similar to how dream skills work. Also, the private skill should only be applied to the particular skill bot and not anywhere else.

Yes

hkedia321:
Fixes #1043 

Changes: 
Enable users to use "private skills" in the SUSI chatbot. The implementation is similar to how "persona" works. 
To apply the private skill for a particular query, the client has to send the following additional parameters:
- `privateskill`
- `userid`
- `group`
- `language`
- `skill`

Then the server fetches the appropriate skill file from `susi_private_skill_data` folder. It then applies the private skill similar to how it applies a dream skill.
Example API: `http://localhost:4000/susi/chat.json?q=hi&privateskill=1&userid=17a70987d09c33e6f56fe05dca6e3d27&group=Knowledge&language=en&skill=knowprides`
The skill exists in the correct location:
image
The skill file content is:
image

Thus, on sending the query "tell me" with the other parameters, we get the correct reply i.e "yes sure" from the server: 
image

### DIFF
--- a/src/ai/susi/server/api/susi/SusiService.java
+++ b/src/ai/susi/server/api/susi/SusiService.java
@@ -6,12 +6,12 @@
  *  modify it under the terms of the GNU Lesser General Public
  *  License as published by the Free Software Foundation; either
  *  version 2.1 of the License, or (at your option) any later version.
- *  
+ *
  *  This library is distributed in the hope that it will be useful,
  *  but WITHOUT ANY WARRANTY; without even the implied warranty of
  *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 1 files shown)


mariobehling:
@Orbiter I think Devs aren't aware of Personas. Could you point Devs to the relevant code please?

Orbiter:
It looks like that personas and private skills require the same function: making a skill a priority over all others == activating it. So we would not new code, just unify the concept. Let me look into detail once I get good internet back...

PR #1119, opened 2018-08-08
Allow private skill to be used only through configured domains

### LINKED ISSUE #1112: Prevent usage of private skill chat from unauthorised domains
**Actual Behaviour**

The API used to fetch responses for a private skill, can be used by anyone from any website. 
Eg. `https://api.susi.ai/susi/chat.json?q=hi&privateskill=1&userid=07628ef8bd403c0c0daef1dc435ad0ed&group=Lifestyle&language=en&skill=My_Portfolio`.

**Expected Behaviour**

If the creator of that private skill has configured the skill such that it should run only on sites specified by the creator, then the above chat API should work only when the request is coming from an authorised site.

Yes.

**Proposed solution:**
On inspecting the header of the AJAX request, I see that the parameter `origin` contains the url of the client website. So I think we can use this to check if the request is coming from an authorised website or not.
image

So we can check the header of the request which contains the `origin` of the request. If the site is included in the list of the sites which the user has approved, then the chat should work otherwise not.

hkedia321:
Fixes #1112 

Changes: 
- In `GetSkillMetaDataService.java` check if the bot creator has selected `allow_bot_only_on_own_sites` option. If `false`, then the API will work from any domain. If `true`, then send response only if the requesting site's domain is included in the `allowed_sites` list. The domain is extracted from the `Referer` field of the request header.
- Similarly in `SusiService.java`, send response only if the the current site is allowed by the bot creator.

These checks are only for the private skills. For other public skills / personas, it will work as usual.

Screenshot for the change:
*Not allowed from other domains:*
(For `getSkillMetaData.json` API)
image
(For `chat.json` API)
image

*Allowed on approved domains:*
(For `getSkillMetaData.json` API)
image
(For `chat.json` API)
image

### DIFF
--- a/src/ai/susi/server/api/susi/SusiService.java
+++ b/src/ai/susi/server/api/susi/SusiService.java
@@ -40,6 +40,7 @@ import ai.susi.json.JsonTray;
 import org.json.JSONTokener;
 
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletRequest;
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.File;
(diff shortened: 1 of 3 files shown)


src/ai/susi/server/api/susi/SusiService.java
```diff
@@ -182,9 +183,37 @@ public ServiceResponse serviceImpl(Query post, HttpServletResponse response, Aut
                         languageName = groupName.getJSONObject(language);
                         if (languageName.has(skill_name)) {
                             skillName = languageName.getJSONObject(skill_name);
-                            if (skillName.getJSONObject("configure").getBoolean("enable_default_skills") == false) {
+                            Boolean allowed_site = true;
+                            JSONObject configureName = skillName.getJSONObject("configure");
+                            if (configureName.getBoolean("enable_default_skills") == false) {
                                 include_default_skills = false;
                             }
+                            if (configureName.getBoolean("allow_bot_only_on_own_sites") && configureName.has("allowed_sites") && configureName.getString("allowed_sites").length() > 0) {
```
Orbiter: This is exactly the same code as in the other class above. We should not do this kind of copy-paste coding because if we must fix a bug somewhere and that is inside such a place, we almost certainly forget to fix the bug in all the affected places.

Please make the functionality of this code a bit abstract, put it into a method and place the method where it belongs most. Then call that method from both servlets.

That is actually a basic thing. Why does this copy-paste coding appear again and again?

hkedia321:
@Orbiter I have now made common method in `DAO.java` file. Please review. Thanks.

Issue #1100, opened 2018-08-02
Implement user avatar upload feature


akshatnitd:
**Actual Behaviour**
* Feature not present

**Expected Behaviour**

- [x] Implement user avatar upload API for users.
- [ ] Implement servlet to get Avatar of a user
- [ ] Send the avatar in login response
- [ ] Implement flag to select between gravatar or user uploaded image

**Steps to reproduce it**

**Would you like to work on the issue?**
Yes

PR #1123, opened 2018-08-11
Ref #1100: Make an API to upload avatar image

akshatnitd:
Ref #1100 

Changes: 
* Make an API to upload avatar image
* It takes 2 params - access_token : String, image : file

### DIFF
--- a/src/ai/susi/server/api/aaa/UploadAvatarService.java
+++ b/src/ai/susi/server/api/aaa/UploadAvatarService.java
@@ -1,0 +1,161 @@
+package ai.susi.server.api.aaa;
+
+import ai.susi.DAO;
+import ai.susi.json.JsonObjectWithDefault;
+import ai.susi.mind.SusiSkill;
+import ai.susi.server.APIHandler;
+import ai.susi.server.AbstractAPIHandler;
+import ai.susi.server.Authentication;
+import ai.susi.server.Authorization;
+import ai.susi.server.ClientCredential;
+import ai.susi.server.ClientIdentity;
+import ai.susi.server.Query;
+import ai.susi.server.ServiceResponse;
+import ai.susi.server.UserRole;
+
+import org.eclipse.jgit.api.Git;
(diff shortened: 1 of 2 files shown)


src/ai/susi/server/api/aaa/UploadAvatarService.java
```diff
@@ -0,0 +1,180 @@
...
+                return token.substring(token.indexOf("=") + 2, token.length()-1);
+            }
+        }
+        return "";
+    }
+
+    /**
+     * Utility method to get MD5 hash of a string
+     */
+    private static String getMD5Hash (String inputText) {
```
Orbiter: Don't implement the MD5 algorithm here because we have that already in Digest.encodeMD5Hex().

Orbiter: furthermore: why do you use the MD5 of the userEmail? We already have an ID which is a hash that is passed in identity.getUuid(). That is in your current variable `userId`. That is the unique identifier for the user, thus that is the most appropriate name of the ìmage_name`.

Orbiter: So I suggest that you omit the MD5 method and the encoding to MD5 completely, just ise the `userId`

akshatnitd: I had encoded the email to MD5 hash, because the same implementation has been done by Gravatar and both the implementation could be kept similar.

akshatnitd: @hkedia321 I think a separate endpoint for uploading avatar should be made. Also, it should be under the `/aaa` folder unlike the `/cms/uploadImage.json` endpoint.

akshatnitd:
@hkedia321 I think a separate endpoint for uploading avatar should be made. Also, it should be under the `/aaa` folder unlike the `/cms/uploadImage.json` endpoint.

PR #1124, opened 2018-08-11
Ref #1100: Make a servlet to Get Avatar Image of user

akshatnitd:
Ref #1100

Changes: 
* Make a servlet to Get Avatar Image of user
* It returns the access token based on the settings of the user, if he wants to use the gravatar image or the server uploaded image.
Endpoint - `http://127.0.0.1:4000/getAvatar.png?access_token=<access-token>`

### DIFF
--- a/src/ai/susi/server/api/aaa/GetAvatarServlet.java
+++ b/src/ai/susi/server/api/aaa/GetAvatarServlet.java
@@ -1,0 +1,112 @@
+/**
+ *  GetAvatarService
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2.1 of the License, or (at your option) any later version.
+ *
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


akshatnitd:
@Orbiter Please review. Also, a default placeholder image has to be added for it, inside the `data/avatar_uploads` folder with image name as `default.jpg`

src/ai/susi/server/api/aaa/GetAvatarServlet.java
```diff
@@ -0,0 +1,86 @@
...
+	private static final long serialVersionUID = 7408901113114682419L;
+
+	@Override
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+
+        Query post = RemoteAccess.evaluate(request);
+        File imageFile = null;
+        String file = "";
+
+        if (post.get("image","") != "") {
```
Orbiter: I don't think that it is good that the client has to know the image file or path. The client should just say "give me the users avatar". You can use the user's credential to extract the Uuid of the user. That is used to store the image according to my latest comment in the corresponding pull request

akshatnitd:
@Orbiter @hkedia321 @AnupKumarPanwar @Akshat-Jain Please review.

PR #1121, opened 2018-08-09
Fixes #1120: Add filter based on staff picks on ListSkillService

akshatnitd:
Fixes #1120 

Changes: [Add here what changes were made in this issue and if possible provide links.]
* Add filter based on staff picks on ListSkillService

PR #1139, opened 2018-08-17
Fixes #1122: Return skill stats to Operator and above role only

### LINKED ISSUE #1122: Return skillStats only to ADMIN and above users
**Actual Behaviour**
* Currently, it is shown to all  USERs

**Expected Behaviour**
* Return skillStats only to ADMIN and above users

Yes

akshatnitd:
Fixes #1122 

* Return skill stats to Operator and above role only


src/ai/susi/server/api/cms/ListSkillService.java
```diff
@@ -358,19 +365,22 @@ else if (filter_type.equals("feedback")) {
...
-        skillStats.put("totalSkills", totalSkills);
-        skillStats.put("reviewedSkills", reviewedSkills);
-        skillStats.put("nonReviewedSkills", nonReviewedSkills);
-        skillStats.put("editableSkills", editableSkills);
-        skillStats.put("nonEditableSkills", nonEditableSkills);
-        skillStats.put("staffPicks", staffPicks);
-        skillStats.put("systemSkills", systemSkills);
+        if(shouldReturnSkillStats) {
+            JSONObject skillStats = new JSONObject();
+            skillStats.put("totalSkills", totalSkills);
```
arundhati24: Should this not be in snake case?

akshatnitd: Yes @arundhati24 , it should ideally be. But, this erroneous naming is used in the entire file and would break the key value used in the different client repositories. Also, I plan to correct the naming in a future PR that would handle all the naming related issues. It seems to be a bit out of scope of the issue.

arundhati24: Okay @akshatnitd 👍

akshatnitd:
@Akshat-Jain The main aim of the PR is to avoid the stats being shown to any Anonymous user or any user with basic rights, as it is an information that we wouldn't like to expose. The PR complies with the use case. I am changing the PR title to match the change.

Akshat-Jain:
@akshatnitd I don't think `REVIEWER` user role should have access to the skill stats. This is in accordance with the fact that `REVIEWER` do not have access to the Admin Panel as well, according to the user role spreadsheet.

Akshat-Jain:
@akshatnitd Can you please open an issue for the resetting issue and work on it on higher priority? It was caused by https://github.com/fossasia/susi_server/pull/1121. Thanks :-)

PR #1146, opened 2018-08-17
Add parameter to exclude default skills in chat API

### LINKED ISSUE #1145: Include/Exclude default susi skills in instant chat
**Actual Behaviour**

In the `instant` chat API (`susi/chat.json?q=hi&instant=hi\nhi%20there`) used for displaying Preview feature on the botbuilder page, there is no way to either include or exclude default susi skills explicitly. The default susi skills are always included.

**Expected Behaviour**

As the user may choose between including / excluding default susi skills, it should reflect in the preview chat. Thus the `/susi/chat.json?instant=` API should accept a parameter to exclude default susi skills.

Yes

Related Issue in client: https://github.com/fossasia/susi_skill_cms/issues/1550

hkedia321:
Fixes #1145 

Changes: 
- If the client provides the parameter `excludeDefaultSkills` with the value `true`, then don't include default SUSI skills.  Any other case (not providing this parameter or any giving any value other than `true`) will result in including default SUSI skills.

Screenshots for the change: 
For API: `http://localhost:4000/susi/chat.json?q=yo&instant=hi\nhi%20there&excludeDefaultSkills=true`, default susi skills are not applied.
image

For API: `http://localhost:4000/susi/chat.json?q=yo&instant=hi\nhi%20there`, default susi skills are applied.
image

### DIFF
--- a/src/ai/susi/server/api/susi/SusiService.java
+++ b/src/ai/susi/server/api/susi/SusiService.java
@@ -88,7 +88,12 @@ public class SusiService extends AbstractAPIHandler implements APIHandler {
         String userId = post.get("userid", "");
         String group_name = post.get("group", "");
         String skill_name = post.get("skill", "");
-        Boolean include_default_skills = true;
+        String excludeDefaultSkills = post.get("excludeDefaultSkills","");
+        Boolean exclude_default_skills = false; // if this is true, default skills would be excluded
+
+        if (excludeDefaultSkills.equalsIgnoreCase("true")) {
+            exclude_default_skills = true;
+        }
 
         try {
             DAO.susi.observe(); // get a database update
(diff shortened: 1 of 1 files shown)


src/ai/susi/server/api/susi/SusiService.java
```diff
@@ -90,6 +90,10 @@ public ServiceResponse serviceImpl(Query post, HttpServletResponse response, Aut
         String skill_name = post.get("skill", "");
         Boolean include_default_skills = true;
 
+        if (post.get("excludeDefaultSkills", null) != null) {
```
Orbiter: This is extremely confusing. What does that mean? If I pass "yes", it has the opposite meaning. Why would an option named "excludeDefaultSkills" should set a variable "excludeDefaultSkills"? With opposite meaning?

Please use one name for options and variables. Explain what the default is. Pass a boolean value to change the meaning. Make the default of the boolean value "false", because that would be natural.

hkedia321:
@Orbiter I have made the changes. Now the main variable is `exclude_default_skills` which is set to `false` by default. If the client provides a parameter `excludeDefaultSkills: true`, then `exclude_default_skills` is set to `true` and the default skills are excluded. In other cases, it is included by default. Please review. Thanks.

Issue #1179, opened 2018-11-02
Same error codes are being used for all API exceptions.


yashasvi128:
**Actual Behaviour**
Same Error Codes are used for all API Exceptions
```
       if (!skill.exists()) {
            throw new APIException(422, "Skill does not exist.");
        }

        if (user_bookmark == null) {
            throw new APIException(422, "Bookmark not provided.");
        }

        skill_bookmark = Integer.parseInt(user_bookmark);

        if(skill_bookmark != 0 && skill_bookmark != 1) {
            throw new APIException(422, "Invalid Bookmark provided.");
        }
```
The 422 (Unprocessable Entity) status code means the server understands the content type of the request entity (hence a 415(Unsupported Media Type) status code is inappropriate. 

**Expected Behaviour**
Each type of error must have a different error code

**Would you like to work on the issue?**
Yes.

ms10398:
As this issue is there for quite some time and no PR is made and I see lots of places using the same error code which is not appropriate.

I am working on the issue and start by changing inappropriate error codes.

@akshatnitd @dynamitechetan is there anything special I need to keep in mind while doing it

PR #1189, opened 2018-11-30
For #1179 : Correct Inappropriate error codes in api/aaa routes

ms10398:
First steps for issues #1179 

Changes: Use appropriate error codes for the API Exception in api/aaa routes

@akshatnitd @Akshat-Jain Please review


Orbiter:
hi @ms10398 good work!

PR #1191, opened 2018-12-03
Fix: Make swagger-ui working again

ms10398:
Changes: Swagger was used on gh-pages but it was broken due to CORS header. I have added CORS proxy to make the request complete and make Swagger working again :slightly_smiling_face:

Screenshots for the change: 
screenshot from 2018-12-03 12-50-47

After the PR is merged and the gh-pages is updated at dev.susi.ai.

The Swagger docs would be able to previewed at https://dev.susi.ai/apidocs/ which is currently broken.

After updation we can update the links in readme for new devs to be redirected at the above link for better understading of the API.

For Testing how will it work

- Go to http://illustrious-bucket.surge.sh/
- Input this url in input box `https://cors.io/?https://api.susi.ai/docs/swagger.json`
- It will serve the doc using the hosted susi-server

@Orbiter @dynamitechetan @akshatnitd Please review

Thanks!

### DIFF
--- a/apidocs/index.html
+++ b/apidocs/index.html
@@ -74,7 +74,7 @@ window.onload = function() {
   
   // Build a system
   const ui = SwaggerUIBundle({
-    url: "http://api.susi.ai/docs/swagger.json",
+    url: "https://api.susi.ai/docs/swagger.json",
     dom_id: '#swagger-ui',
     deepLinking: true,
     presets: [


apidocs/index.html
```diff
@@ -74,7 +74,7 @@
   
   // Build a system
   const ui = SwaggerUIBundle({
-    url: "http://api.susi.ai/docs/swagger.json",
```
Orbiter: I debugged the `https://dev.susi.ai/apidocs/` page and found out that the error message was "mixed content". So this does not complain about CORS but only that the endpoint is http, not https.

Please use the url `https://api.susi.ai/docs/swagger.json` (added 's') instead the cors.io endpoint as a first try, so we depend not on an external service

PR #1196, opened 2018-12-07
Refactor to use better syntax and correct semantic errors

ms10398:
For #1195 

Changes: Detailed explanation has been added to file changes using comments

@Orbiter please review and check if nothing is broken.

PR #1198, opened 2018-12-08
Changed variables to local which can be used as a local variable instead of keeping them global

ms10398:
For #1196 

Changes: Comments have been added to explain the changes.


conf/os_skills/system/en/en_0301_bahn.json
```diff
@@ -5,7 +5,7 @@
 		],
 		"process":[ {"type":"console", "expression":"SELECT description FROM bahn WHERE from='$1$' to='$2$';"}],
 		"actions":[ {"type":"answer", "select":"random", "phrases":[
-			"$description$",
```
ms10398: According to JSON specification there should not be trailing commas present in JSON

Reference - https://stackoverflow.com/a/201856

src/ai/susi/DAO.java
```diff
@@ -98,8 +98,8 @@
     public final static int ATTENTION_TIME = 5;
     private final static String ACCESS_DUMP_FILE_PREFIX = "access_";
     public  static File conf_dir, bin_dir, html_dir, data_dir, skill_status_dir, susi_chatlog_dir, susi_skilllog_dir, draft_dir, model_watch_dir, susi_skill_repo, private_skill_watch_dir, susi_private_skill_repo, deleted_skill_dir, system_keys;
-    private static File external_data, assets, dictionaries;
```
ms10398: Converted field to local variable as it is only used locally

ms10398:
I have changed variables scope from global to local where there was no need to keep them as global variable.

The build passes successfully with the test :slightly_smiling_face:

PR #1199, opened 2018-12-12
Fixes #1179: Update Error codes

### LINKED ISSUE #1179: Same error codes are being used for all API exceptions.

sakshee-19:
References issue #1179

Changes: Updated error codes and messages for existing error codes.

Screenshots for the change:

### DIFF
--- a/src/ai/susi/server/api/aaa/ChangeUserRoles.java
+++ b/src/ai/susi/server/api/aaa/ChangeUserRoles.java
@@ -149,7 +149,7 @@ public class ChangeUserRoles extends AbstractAPIHandler implements APIHandler {
             ClientCredential credential = new ClientCredential(ClientCredential.Type.passwd_login, userTobeUpgraded);
             ClientIdentity identity = new ClientIdentity(ClientIdentity.Type.email, credential.getName());
             if (!DAO.hasAuthorization(identity)) {
-                throw new APIException(400, "Username not found");
+                throw new APIException(404, "Username not found");
             }
 
             // Update the local database
@@ -157,7 +157,7 @@ public class ChangeUserRoles extends AbstractAPIHandler implements APIHandler {
             try {
                 auth.setUserRole(userRole);
             } catch (IllegalArgumentException e) {
-                throw new APIException(400, "role not found");
+                throw new APIException(404, "role not found");
             }
 
             // Print Response
--- a/src/ai/susi/server/api/aaa/DeleteGroupService.java
+++ b/src/ai/susi/server/api/aaa/DeleteGroupService.java
@@ -81,7 +81,7 @@ public class DeleteGroupService extends AbstractAPIHandler implements APIHandler
                 result.put("message", "Group deleted successfully");
                 return new ServiceResponse(result);
             } else {
-                throw new APIException(422, "Group doesn't exists");
+                throw new APIException(401, "Group doesn't exists");
             }
         } else {
             throw new APIException(400, "Group name parameter not specified");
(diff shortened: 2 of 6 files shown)


src/ai/susi/server/api/aaa/ChangeUserRoles.java
```diff
@@ -149,15 +149,15 @@ public ServiceResponse serviceImpl(Query query, HttpServletResponse response, Au
...
-                throw new APIException(400, "Username not found");
+                throw new APIException(404, "Username not found");
             }
 
             // Update the local database
             Authorization auth = DAO.getAuthorization(identity);
             try {
                 auth.setUserRole(userRole);
             } catch (IllegalArgumentException e) {
-                throw new APIException(400, "role not found");
```
ms10398: @Orbiter just to be sure 

But the exception was thrown because the query string didn't had proper role argument so instead of a `404` it should be a `400` -> Bad Data

What do you suggest about it

src/ai/susi/server/api/aaa/LoginService.java
```diff
@@ -298,9 +298,9 @@ else if(pubkeyLogin){ // second part of pubkey login: verify if the response to
 				sig.update(challenge.getBytes());
 				verified = sig.verify(Base64.getDecoder().decode(challangeResponse));
 			} catch (NoSuchAlgorithmException e){
-				throw new APIException(400, "No such algorithm");
```
ms10398: I think here should also not be a 404 because the parameter are wrong but the route exist

src/ai/susi/server/api/aaa/ModifyUserDevices.java
```diff
@@ -98,7 +98,7 @@ public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Aut
                             }
                         }
                         else {
-                            throw new APIException(400, "Specified device does not exist.");
```
ms10398: I think it should also not be a 404

As the device name which was provided as query string is not present

So the error code should be `422` or `400`

src/ai/susi/server/api/aaa/DeleteGroupService.java
```diff
@@ -81,7 +81,7 @@ public ServiceResponse serviceImpl(Query post, HttpServletResponse response, Aut
                 result.put("message", "Group deleted successfully");
                 return new ServiceResponse(result);
             } else {
-                throw new APIException(422, "Group doesn't exists");
+                throw new APIException(401, "Group doesn't exists");
```
ms10398: 401 is used for `unauthorized` status

but here the group doesn't exist so shouldn't it be a `422` or `400`

src/ai/susi/server/api/aaa/ChangeUserRoles.java
```diff
@@ -149,15 +149,15 @@ public ServiceResponse serviceImpl(Query query, HttpServletResponse response, Au
             ClientCredential credential = new ClientCredential(ClientCredential.Type.passwd_login, userTobeUpgraded);
             ClientIdentity identity = new ClientIdentity(ClientIdentity.Type.email, credential.getName());
             if (!DAO.hasAuthorization(identity)) {
-                throw new APIException(400, "Username not found");
+                throw new APIException(404, "Username not found");
```
ms10398: It shouldn't be a 404 as the route is present and a param of username provided is only wrong

PR #1229, opened 2019-02-03
Fixed:#1228 new feature to strengthen user Passwords

### LINKED ISSUE #1228: Feature to prevent the use of simple "Dictionary" words as the Login Password.
**Actual Behaviour**

While resetting the login password, the new password String might be a simple dictionary word/noun. Such set of passwords should be discarded if we are to prevent various "dictionary attacks" on the user's account. The SUSI server currently considers such combinations valid. 
The system easily accepts such weak passwords which are prone to various attacks.

**Expected Behaviour**

This feature will not allow the users to register predictable dictionary words as their Login passwords. A toast message informing the vulnerability of such passwords will be shown. Currently this added feature will only work on systems running the LINUX Operating System.

**Steps to reproduce it**

**LogCat for the issue**

**Screenshots of the issue**

Yes

dernDren161:
Fixes #1228 

Changes: The suggested feature is added. 

Screenshots for the change:


Orbiter:
This commit contains 32 files of changes for class files and other rubbish that should not be part of a pull request for one single file. This cannot be merged.

PR #1238, opened 2019-02-23
#Fixes:#1235 updated file to render correct response message

### LINKED ISSUE #1235: Render the correct response message when the server is hosted on "localhost".
**Actual Behaviour**

When running the server on a local machine and hosting it on the "localhost", faulty messages are being thrown after the completion of certain tasks.
Specifically, after successfully changing the password, we get a wrong APIException message as: "Failed to Change the Password" despite the password being successfully updated. 
This is primarily because the PasswordChangeService.java route directly jumps to the catch statement when running on the localhost and hence shows the wrong message. The current code structure works perfectly fine otherwise but malfunctions when is tested locally.  

**Expected Behaviour**

No matter where the endpoint is used, it should be able to give uniform and correct response messages. Particularly, the PasswordChangeService.java file should give a "Success" message after the successful change in password when run on the localhost.

**Steps to reproduce it**

**LogCat for the issue**

**Screenshots of the issue**
This is being displayed though the password has successfully been changed. 
error 

Yes

dernDren161:
Fixes #1235 

Changes: The changes in the file now helps to render the correct response messages.

Screenshots for the change: 
screenshot from 2019-02-23 10-07-44

### DIFF
--- a/src/ai/susi/server/api/aaa/PasswordChangeService.java
+++ b/src/ai/susi/server/api/aaa/PasswordChangeService.java
@@ -23,6 +23,9 @@ import ai.susi.tools.TimeoutMatcher;
 import org.json.JSONObject;
 
 import javax.servlet.http.HttpServletResponse;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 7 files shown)


src/ai/susi/server/api/aaa/PasswordChangeService.java
```diff
@@ -113,18 +116,29 @@ public ServiceResponse serviceImpl(Query post, HttpServletResponse response, Aut
...
                 try {
                     EmailHandler.sendEmail(authentication.getIdentity().getName(), subject, "Your password has been changed successfully!");
                     result.put("message", "Your password has been changed!");
                     result.put("accepted", true);
                 } catch (Exception e) {
-                    throw new APIException(500, "Failed to change password");
+                    if(findIP.equals(IP)) {
+                    	result.put("message","Your password has sucessfully been changed and stored in the local database!");
+                    	result.put("accepted",true);
+                }
```
Orbiter: please keep the Excpetion in the else case so we still have a hint that something did not work

PR #1239, opened 2019-02-24
Fixes:#1235 render correct messages and  exceptions

### LINKED ISSUE #1235: Render the correct response message when the server is hosted on "localhost".

dernDren161:
Fixes #1235 
Changes: A revamped solution where there is also the inclusion of an "Exception" which gives an error message if in case the local server is not able to update the password due to a "Server error".
Screenshots for the change:

### DIFF
--- a/src/ai/susi/server/api/aaa/PasswordChangeService.java
+++ b/src/ai/susi/server/api/aaa/PasswordChangeService.java
@@ -23,6 +23,9 @@ import ai.susi.tools.TimeoutMatcher;
 import org.json.JSONObject;
 
 import javax.servlet.http.HttpServletResponse;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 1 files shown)


dernDren161:
@Orbiter  Sir, the commit is updated as per your guidance. Apart from this i also have a couple of doubts in this file:
1) the use of string "newsalt" is inevident in the entire file and has no use, i reckon it was created to be included with the "newpassword" but "newpassword" uses the same old "salt", so sir either we have to delete the "newsalt" or use it with the "newpassword". Which one should i do if i am to refactor the code?
2)when changing the password locally the password in the SUSI server account is not updated. Should changes be made in order to update the server password?
3) while testing the local code, an unauthorized user with any invalid email address can easily get the "access_token" which I think should be changed.
Thank You

PR #1250, opened 2019-04-20
fix

foradamatrix:

Screenshots for the change:


Orbiter:
It would help to process such pull requests if the description would not only contain unedited template text.

norbusan:
In addition it is a mixture of stuff that is already applied and something else. This branch should be cleaned up and rebased on development.

PR #1251, opened 2019-04-24
add systemd service file and install into release tarball

norbusan:
This service file is a service template, so can be started as an arbitrary user using
`systemctl start ss-susi-server@USER`.

Furthermore, the file needs configuration of `@INSTALL_DIR@` to set the actual installation place.

install anything in the `systemd` directory also into the release tarball


norbusan:
Thanks

PR #1254, opened 2019-05-08
allow for logging to stdout while still daemonizing

norbusan:
In the systemd service unit we want the daemon be of type
forking, so that systemd can reliable determine whether startup
has completed.

We add a setting SS_LOG_TO_STDOUT (cmd line option -o) to start.sh
that triggers this.

For the systemd unit we set the above env variable to 1, set
the type to forking, and in addition specify for stability the
location of the PID file.

### DIFF
--- a/bin/start.sh
+++ b/bin/start.sh
@@ -8,7 +8,11 @@ cd $(dirname $0)/..
 # Execute preload script
 source bin/preload.sh
 
-while getopts ":Idn" opt; do
+#
+# default log to file, but take env var into account
+SS_LOG_TO_STDOUT=${SS_LOG_TO_STDOUT:-0}
+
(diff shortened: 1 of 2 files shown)


norbusan:
Well, this is a question of the log4j configuration file. Either we update the confg file, or we add `>&2` to the `$cmdline`.

norbusan:
@hongquan  BTW, something is strange with either my computer or yours, because it shows my
```
 hongquan commented 7 hours from now
```
Mind the **from now** so in the future ...

hongquan:
@norbusan It is weird that logging is thrown to `stdout`, instead of `stderr`. I feel it is anti-pattern.

hongquan:
> Mind the from now so in the future ...

@norbusan Haha

PR #1279, opened 2019-06-21
fix JARFILE name

### LINKED ISSUE #1278: bin/start.sh doesn't start the server after successful build
**Actual Behaviour**
image

- The JARFILE name in `bin/preload.sh` doesn't match the generated one by the updated gradle wrapper

**Steps to reproduce it**
- build the project
- run `bin/start.sh`

yes

alok760:
Fixes #1278

Changes: update bin/preload.sh with the correct JARFILE name

Screenshots for the change: 
image

### DIFF
--- a/bin/preload.sh
+++ b/bin/preload.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-JARFILE="build/libs/susi_server-all.jar"
+JARFILE="build/libs/susi_server.jar"
 INSTALLATIONCONFIG="data/settings/installation.txt"
 PIDFILE="data/susi.pid"
 DFAULTCONFIG="conf/config.properties"


bin/preload.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-JARFILE="build/libs/susi_server-all.jar"
+JARFILE="build/libs/susi_server.jar"
```
norbusan: No, I don't think this is a good idea. We later need to deploy the stuff and thus the `-all` variant needs to be build. The correct incantation is `gradle assemble`.

I think it should be only documented that `gradle assemble` is necessary.

alok760: aah ok, Thankyou @norbusan 
closed the PR

PR #1288, opened 2019-07-01
Fix Loading of List User in Admin

### LINKED ISSUE #1287: Fix Loading of List User in Admin
**Actual Behaviour**
Currently the users loaded in admin panel.

**Expected Behaviour**
Fix it.

Yes

harsh-jindal:
Fixes #1287 

Changes: Made changes in aaa/GetUsers.java

### DIFF
--- a/src/ai/susi/server/api/aaa/GetUsers.java
+++ b/src/ai/susi/server/api/aaa/GetUsers.java
@@ -46,13 +46,6 @@ public class GetUsers extends AbstractAPIHandler implements APIHandler {
 
     @Override
     public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Authorization rights, final JsonObjectWithDefault permissions) throws APIException {
-        if (call.get("getPageCount", false) == false
-                && call.get("getUserStats", false) == false
-                && call.get("search", null) == null
-                && call.get("page", null) == null
-                && call.get("getUserCount", null) == null) {
-            throw new APIException(400, "Bad Request. No parameter present");
-        }
         JSONObject result = new JSONObject(true);
         JSONObject userStats = new JSONObject(true);
         result.put("accepted", false);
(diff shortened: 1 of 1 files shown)


harsh-jindal:
@Orbiter @akshatnitd @fragm3 Please review.

fragm3:
I tried to make it work, a few days back. But the problem is, it's working on localhost, without any changes. How can we test if it can work on a real server instance? @akshatnitd

harsh-jindal:
@fragm3 It doesn't work for me on my localhost. Also it's strange how it works for you because if you pass only access token as parameter you shall get `Bad request. No parameter present` as per the current code. Also you get java.array.outofbound exception because of page =0.

fragm3:
@harsh-jindal I tried with correct query parameters, which are specified in the Servlet file.

harsh-jindal:
@fragm3 Then it would be best if we have some way to check it in on real server instance as said by you.

src/ai/susi/server/api/aaa/GetUsers.java
```diff
@@ -46,13 +46,6 @@ public JSONObject getDefaultPermissions(UserRole baseUserRole) {
 
     @Override
     public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Authorization rights, final JsonObjectWithDefault permissions) throws APIException {
-        if (call.get("getPageCount", false) == false
```
akshatnitd: This check is important. Do not remove it.

src/ai/susi/server/api/aaa/GetUsers.java
```diff
@@ -74,7 +67,7 @@ public ServiceResponse serviceImpl(Query call, HttpServletResponse response, Aut
             result.put("message", "Success: Fetched count of users");
             return new ServiceResponse(result);
         } else {
-            int page = call.get("page", 0);
+            int page = call.get("page", 1);
```
akshatnitd: Default page should be 0

PR #1324, opened 2019-07-21
Fix password RegEx #1323

### LINKED ISSUE #1323: Fix password RegEx
**Actual Behaviour**

Currently, password regex requires an upper case letter as well

**Expected Behaviour**

Require user passwords to have at least 8 characters and at least one special character, one number, and text characters

Needed for PR: https://github.com/fossasia/susi.ai/pull/2639

Yes

fragm3:
Fixes #1323

Changes: Currently, password regex requires an upper case letter as well. Require user passwords to have at least 8 characters and at least one special character, one number, and text characters


fragm3:
@akshatnitd please review

PR #1351, opened 2019-07-30
Modify getUsers to return device Stats and fix getUserStats

### LINKED ISSUE #1350: Modify getUsers to return device Stats
**Actual Behaviour**
Not preset. Required for https://github.com/fossasia/susi.ai/issues/2706

**Expected Behaviour**
Implement it.

Yes

harsh-jindal:
Fixes #1350 
Changes: `/aaa/getUsers.json?getDeviceStats=true`returns  number of connected devices and device users.
- [x] - Number of Connected devices 
- [x] - Number of device users 
- [x] - device added over time
Screenshots for the change: 
Screenshot 2019-07-31 at 11 09 51 PM

### DIFF
--- a/src/ai/susi/server/api/aaa/AddNewDevice.java
+++ b/src/ai/susi/server/api/aaa/AddNewDevice.java
@@ -41,6 +41,8 @@ import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
+import java.util.Date;
+import ai.susi.tools.DateParser;
 
 
 /**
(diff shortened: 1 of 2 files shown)


harsh-jindal:
@fragm3 @akshatnitd @mariobehling If we want the number of devices deleted by the user as it is required for total number of deleted devices. The approach I was thinking is to add a prop `deletedDevices` when the user connects his first device in the file `AddNewDevice` and whenever a device is deleted by the user or admin increment it by one. Is this approach correct ?

harsh-jindal:
@fragm @akshatnitd Please review

harsh-jindal:
@akshatnitd Done.Made the changes. Also fixed getUserStats in this PR.
@fragm3 As you had suggested last time that on the online server `substring(8,16)` would not give the correct result but thats not the case. For skill creation we have a different parser compared to login, signup, and device added. That is why for skill in `getSkillList.json` we use `substring(0,8) `

PR #1444, opened 2020-01-30
Fixed !!

whitedevil4:

Screenshots for the change:


norbusan:
@Orbiter ok, so we have updated the master branch. Does that mean you consider the current state good enough concerning memory management etc?

PR #1458, opened 2020-02-14
Resolved store draft issue

### LINKED ISSUE #1457: StoreDraftService not able to save any draft
**Actual Behaviour**

Currently the endpoint returns
```
object parser error: Unterminated string at 313 [character 314 line 1]
```

**Expected Behaviour**

Resolve it so that drafts can be saved.

Yes

Dilshaad21:
Fixes #1457 

Changes: 
- Accepted all query params separately and stored it in `object`

Screenshots for the change: 
NA


Dilshaad21:
Earlier there was parsing issue of the stringified json data. Now I have handled each of the key-value pairs separately.

Dilshaad21:
@Orbiter Changes to made after merging this PR in client-side.
- Instead of passing stringified json, need to pass the object itself.
- Request need to be `POST` as the payload is huge

PR #1469, opened 2020-05-13
chore: Include JUnit tests in CI test phase

### LINKED ISSUE #1466: Travis is not really testing the build
The test script in Travis just runs assemble and clean, and tests are run in `after_success`, hence even if the tests fail, which is the case in latest build, https://travis-ci.org/github/fossasia/susi_server/jobs/680731115, the build will succeed, making the tests useless

Also, for some reason, codecov is run in parallel and obviously can't find the coverage report and errors out. It should be run after the tests

iamareebjamal:
Fixes #1466

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,3 +56,8 @@ matrix:
           script: bash scripts/deploy_hcloud.sh
           on:
             all_branches: true
+
+branches:
+  only:
+    - master
+    - development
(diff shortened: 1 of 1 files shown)


iamareebjamal:
To be clear, this is showing a decrease in coverage because it is comparing it with a diff from 2 years ago.

PR #1473, opened 2020-07-30
CODE_OF_CONDUCT

Gaushikmr:


norbusan:
I strongly oppose to use the "Contributor Covenant Code of Conduct", not for it's content, but for its political background.

I strongly suggest https://github.com/domgetter/NCoC as CoC instead.

PR #1474, opened 2020-08-16
Update README with JDK8 installation and configuration steps for mac.

blossom2017:

Updates the readme with JDK 8 installation and configuration steps for mac users.
Screenshots for the change:


norbusan:
This is very specific to using homebrew while ignoring other ways to install Java on MacOS (macports, direct package, ...), thus, this is not really helpful as is.

Furthermore, I don't think that explanation how to manage multple Java versions using `jenv` are useful here, we don't do it either for Windows nor Linux.

blossom2017:
Ok closing it.

PR #1485, opened 2026-02-19
Replace fixed CI waits with Kubernetes rollout-based deployment checks

### LINKED ISSUE #1470: Builds run for more than 30 minutes
Due to https://github.com/fossasia/susi_server/commit/edf4a80fe5100d91474c4654adb9e78e006f416a, builds run more than 30 minutes where they could run in 3-5 minutes. This wastes CI resources and no PR build can run in that period. There must be a better way to wait for some resource to get ready than to wait for 30 or 10 minutes. Also, the 30 minute wait does nothing because Travis basically kills the script after 30 minutes anyway, so the deployment gets cancelled

MerwinJoshwa:
Fixes #1470

Changes:
- Removed reliance on long static CI waits during deployment.
- Improved deployment logic to rely on Kubernetes readiness instead of fixed time delays.
- Reduced unnecessary CI blocking that caused builds to run for ~30 minutes.

This change prevents Travis builds from being blocked while waiting for resources that are already handled by Kubernetes.

## Summary by Sourcery

Enhancements:
- Add a Kubernetes rollout status wait loop with timeout and diagnostics to the Travis deployment script to avoid long static delays.

### DIFF
--- a/kubernetes/travis/deploy.sh
+++ b/kubernetes/travis/deploy.sh
@@ -38,4 +38,26 @@ echo ">>> Pushing docker image"
 docker push fossasia/susi_server
 
 echo ">>> Updating deployment"
-kubectl set image deployment/susi-server --namespace=web susi-server=fossasia/susi_server:$TRAVIS_COMMIT
+kubectl set image deployment/susi-server \ --namespace=web \susi-server=fossasia/susi_server:$TRAVIS_COMMIT
+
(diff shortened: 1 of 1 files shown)


MerwinJoshwa:
Addresses the root cause of prolonged Travis CI runtimes by replacing fixed wait-based deployment behavior with deterministic execution flow.

